Issue #28, opened 2014-11-25
Allow `source activate` to work without an environment when an environment.yml file is present


tswicegood:
The idea is to check to do something like `conda env info --name` that would read the environment file then output the name.  This could be used inside `source activate` to supply the environment name instead of simply errorring out.

malev:
I think think this is happening anymore

PR #29, opened 2014-11-25
Add first rough pass at fixing #28

tswicegood:
Add in support for doing `source activate` inside a directory with an `environment.yml` file and having it pick up that environment and activate it.

## Testing
In order to test this branch, you have to install conda from this branch: https://github.com/conda/conda/pull/1013.  From within a checkout of that branch, run:

    python setup.py develop

## Tasks
* [x] Add `source activate` support on *nix systems
* [ ] Handle situation where an environment has not been created
* [ ] Add `activate` support on Windows
* [ ] Verify all `test_activate` scripts in conda still pass
* [ ] Document usage

### DIFF
--- a/bin/env-activate
+++ b/bin/env-activate
@@ -36,7 +36,18 @@ get_dirname() {
 }
 source "$(get_dirname ${_ABS_SCRIPT_LOCATION})/_conda-functions"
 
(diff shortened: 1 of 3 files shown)


malev:
I don't think we need this anymore

Issue #40, opened 2014-12-16
conda env update


chdoig:
When working with conda env, I've sometimes encountered that there are new packages in my environment.yml file and I want to update my environment with those. It would be nice to have a `conda env update` command that could do that for you, without having to remove and create a new one.

tswicegood:
This is now released as of 1.4.0 thanks to @aterrel.

PR #41, opened 2015-01-06
Adds an update command

aterrel:
Frequently, someone updates the environment which requires me to go figure out what was added. This command makes conda env able to update an environment based on the current environment.yml.


tswicegood:
Thanks, @aterrel!  Let me try it out and I'll get it merged/released.

Fixes #40.

tswicegood:
Works like a charm.  Plenty of room for refactoring, but it gets the feature in.

Issue #27, opened 2014-11-21
Allow multiple types of lookups when trying to find an environment


tswicegood:
Currently it only looks for a local file in the current working directory or explicitly via `--file`.  Would be nice to allow plugins to add other lookups such as GitHub, URLs, Binstar, and so on.

malev:
Already done!

PR #44, opened 2015-01-16
Feature/remote environments

srossross:
# DO NOT MERGE!

This PR points to **master**

## TODO:

Some TODOs are referenced in the code:

 * [ ] Merge into develop instead of master
 *  [ ] Add tests
 *  [x] Test the the correct error shows up when binstar is not installed
 *  [ ] There are numerous functions in binstar parts of binstar that were originally internal. We need refactor binstar so that individual arguments are passed in instead of an arg object. And make general cli utilities.
 *  [ ] ~~Binstar requires the explicit creation of a version namespace. This should be removed as a hard requirement of binstar.~~ *Don't think we can do this here*
 *  [ ] Check and filter versions from a command line specification 
 *  [ ] ~~Check MD5 of the environment yaml file before fetching it again~~
 *  [ ] Conda should be refactored to include a `Specs` object (or this needs to change to use the conda internals)
 * [ ] ~~Update the environment object to take `version`  and `summary` arguments~~ We are not taking versions for now

* [x] Create another PR with https://github.com/conda/conda-env/commit/0c6706a02b721ac20b94a0e1abe14dfe542a5944


conda_env/cli/main_create.py
```diff
@@ -41,13 +46,17 @@ def configure_parser(sub_parsers):
...
 
+
     p.add_argument(
         '-f', '--file',
-        action='store',
         help='environment definition (default: environment.yml)',
         default='environment.yml',
     )
     p.add_argument(
+        '--binstar',
```
tswicegood: Using a `--binstar` isn't the correct way to do this.  As outlined in #27, the idea is to allow multiple lookups.  I'm using [hub][] as the inspiration.  The UI would then be:

    conda create env tswicegood/stats
    conda create env sean/geo

Both of these are probably outside the scope of this PR though, so I'd say keep that all in a PR related to #27.

[hub]: https://github.com/github/hub/

conda_env/utils/binstar_uilts.py
```diff
@@ -0,0 +1,107 @@
+from binstar_client import errors
```
tswicegood: Typo in the file name.  Should also probably just be `utils/binstar.py` to avoid being redundant.

conda_env/exceptions.py
```diff
@@ -2,6 +2,11 @@ class CondaEnvException(Exception):
     pass
 
 
+class NoBinstar(CondaEnvException):
+    def __init__(self):
+        msg = 'The binstar must be installed to perform this action'
```
malev: "The binstar" sounds so weird!

conda_env/utils/binstar.py
```diff
@@ -0,0 +1,107 @@
...
+            interactive_login(binstar_args(None, None))
+
+            # TODO: refactor binstar to make this easier e.g `binstar.reload_token() or binstar.ensure_login()`
+            # In binstar cli the binstar object is re-created.
+            binstar.token = get_binstar().token
+            binstar._session.headers.update({'Authorization': 'token %s' % (binstar.token)})
+
+            return binstar.user()
+        else:
+            raise
```
malev: What is this empty raise suppose to do? <- @srossross

conda_env/utils/binstar.py
```diff
@@ -0,0 +1,107 @@
...
+
+def ensure_package_namespace(binstar, username, packagename, version=None, summary=None):
```
malev: I'm not a fan of functions that does not returns something

tswicegood: Agreed.  My preference is that a function return what needs to be done, not perform that action on the argument.  There's no immediate feedback that anything happened to `binstar`.

Given the way this function is meant to be used, it seems as though it should be:

```python
if not binstar.package_exists(username, packagename):
    binstar.add_package(username, packagename, version, summary)
```

That seems as though it would require some changes to `binstar_client` to make that work.  That's a bit out-of-scope. :-/

conda_env/utils/binstar.py
tswicegood: The empty raise here works to re-raise the exception.  The idea (correct me if I'm wrong @srossross), is that this gives the user a chance to log in if they aren't already.  If they are and they're still not authorized we don't actually want to catch this.

FWIW, I think this is a common enough case to have an `errors.NotLoggedIn` that is thrown when a user is unauthorized *and* not logged in.  That would remove the need for this `if/else`.

conda_env/cli/main_create.py
```diff
@@ -11,6 +11,10 @@
 from ..env import from_file
 from ..installers.base import get_installer, InvalidInstaller
 from .. import exceptions
+try:
+    from conda_env.utils import binstar as binstar_utils
+except:
```
tswicegood: We shouldn't catch all exceptions here.  This should be `except ImportError`.

tswicegood:
@malev if you're taking a stab at this, do you mind separating out the changes I mentioned [in this comment][1] into a separate PR?  It'd be awesome to see both of them land soon'ish, but I don't want to muddy this up with one PR to do both as I'd like to do GitHub-based remote environments too in that one.

[1]: https://github.com/conda/conda-env/pull/44/files#r23129751

conda_env/utils/binstar.py
malev: The problem with an empty **raise** is that we are showing the
```
An unexpected error has occurred, please consider sending the
following traceback to the conda GitHub issue tracker at:
```
We are blaming our self!! Why not adding a more clear error message about what's going on?

tswicegood: +1 to a more clear error message.  Actually, if you raise an exception that extends from `RuntimeError`, that will be formatted and output [inside conda][1].  Let's do that.  There's a `conda_env.exceptions` module with tests around what should happen.  Add one to that that extends from `CondaEnvException` and `RuntimeError` and you're set.

[1]: https://github.com/conda/conda/blob/master/conda/cli/main.py#L212-L213

tswicegood:
One general thought I want to capture on the PR (it's been offered directly out-of-band).  There's a pretty big leakage of binstar functionality into conda-env to make this work.  Adding helpers inside `conda_env.utils.binstar` is covering functionality cases that should probably be covered in binstar.

For example, `ensure_package_namespace` provides all of the functionality around creating a package if it doesn't exist.  There are two ways to solve this implementation leak:

* Wrap this function inside the binstar API module so it does the various things it needs to
* Add this functionality to the REST-API so there is an endpoint that provides that same functionality

I lean toward the latter as it will scale further, though starting the the API module at least protects external packages such as `conda-env` from the knowledge of what's happening internally.  `conda-env` shouldn't have to know anything about the internals of binstar.  Reaching through for more than upload, download, and create leaks too much implementation into this library.

The solution for now is to isolate all of this functionality inside the `conda_env.utils.binstar` to isolate the rest of the module from knowing anything about the implementation of `binstar`'s module.  This module should eventually be simple imports.

malev:
@tswicegood agree and that's exactly what I am doing now

malev:
@tswicegood I made some changes, every binstar-client related artifact is hidden inside the Uploader class. Please tell me what you think

conda_env/exceptions.py
tswicegood: FWIW, I think it's mean to be "The binstar package" or "The binstar client". :-)

conda_env/utils/uploader.py
```diff
@@ -0,0 +1,117 @@
...
+
+class Uploader(object):
+    """
+    Upload environments to Binstar
+    """
+
+    user = None
+
+    @staticmethod
+    def is_installed():
```
tswicegood: Style comment: I would keep this as a function.  `conda_env.utils.binstar.is_installed` seems like a good place.

malev: Sorry, too much Ruby in my veins

tswicegood: :-)

malev:
This PR has been decomposed into:
* https://github.com/conda/conda-env/pull/78
* https://github.com/conda/conda-env/pull/73
Both of them has been merged therefore I'm closing this one.

PR #45, opened 2015-01-22
v1.4.2 release: fix oddball BASH issue

tswicegood:
From commit log:

> I've got an issue, probably related to my highly customized bash prompt, where $0 == "-/bin/bash".  That cases errors when passed to `basename` because thinks that's a parameter and tries to treat it as such.  This modifies the code to ignore those errors.  The path that its attempting to detect is a direct call to "activate" or "deactivate" which won't trigger those errors.

### DIFF
--- a/bin/env-activate
+++ b/bin/env-activate
@@ -1,8 +1,9 @@
 #!/bin/bash
 
 # Ensure that this script is sourced, not executed
-# Note that if the script was executed, we're running inside bash!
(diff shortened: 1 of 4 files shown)


tswicegood:
Will merge and release as soon as build is successful.

tswicegood:
Hmm -- this succeeded on Binstar, but Binstar hasn't updated GH.  All tests are still passing locally and the builds were all successful, going to merge.

PR #46, opened 2015-01-22
Remove runtime dependency on setuptools

asmeurer:
To install an installer plugin, one simply needs to install a .py file with
the desired name to site-packages/conda_env/installers/

I didn't change any documentation yet (if there is any).

### DIFF
--- a/conda_env/installers/base.py
+++ b/conda_env/installers/base.py
@@ -1,6 +1,4 @@
-import pkg_resources
-
-ENTRY_POINT = 'conda.env.installers'
+ENTRY_POINT = 'conda_env.installers'
 
 
 class InvalidInstaller(Exception):
@@ -10,8 +8,7 @@ class InvalidInstaller(Exception):
 
 
 def get_installer(name):
-    for entry_point in pkg_resources.iter_entry_points(ENTRY_POINT):
-        if entry_point.name == name:
-            return entry_point.load()
-
-    raise InvalidInstaller(name)
+    try:
+        return __import__(ENTRY_POINT + '.' + name)
+    except ImportError:
+        raise InvalidInstaller(name)
--- a/conda.recipe/meta.yaml
+++ b/conda.recipe/meta.yaml
@@ -16,7 +16,6 @@ requirements:
     - setuptools
   run:
     - python
-    - setuptools
 
 test:
   commands:


conda_env/installers/base.py
```diff
@@ -1,6 +1,4 @@
-import pkg_resources
-
-ENTRY_POINT = 'conda.env.installers'
+ENTRY_POINT = 'conda_env.installers'
```
asmeurer: I'm assuming this was a typo

tswicegood:
I'm currently away from my computer. Will take a look at this in a few hours when I'm back online.

teoliphant:
We are stalled on merging PR 1013 until you look at this.

conda_env/installers/base.py
tswicegood: Nope -- it was not a type -- the entry point was based on what the eventual name was supposed to be.  The idea is that eventually `conda_env` would become `conda.env`.

tswicegood:
My only problem with this is that it forces developers to install packages into an existing namespace, something that's fragile at best (that statement is coming from having built out a 30 module tool that used namespaces exclusively to put them all in the same space).

I think we need to eventually create this style of plug-in architecture and I think `setuptools` is going to be the right tool (unless we want to move conda to Python 3.4 only -- as noted in [this discussion][1]) as it allows the most flexibility without forcing users into something that we know is worst than setuptools.

---
All that said, since this is an undocumented feature, I'm ok with the general theme of this PR.  There are some changes that need to be made to ensure it builds properly, but I can deal with that.

[1]: https://github.com/conda/conda/pull/993

teoliphant:
We can revisit making setuptools a dependency when we discuss conda later.   For now, forcing a setuptools dependency at the same time as PR 1013 is not what I want to do.

tswicegood:
This code doesn't appear to run:

```
❯ conda env create
An unexpected error has occurred, please consider sending the
following traceback to the conda GitHub issue tracker at:

    https://github.com/conda/conda/issues

Include the output of the command 'conda info' in your report.


Traceback (most recent call last):
  File "/Users/travis/anaconda/bin/conda-env", line 7, in <module>
    execfile(__file__)
  File "/Users/travis/Dropbox/work/github/conda/conda-env/bin/conda-env", line 5, in <module>
    sys.exit(main())
  File "/Users/travis/Dropbox/work/github/conda/conda-env/conda_env/cli/main.py", line 53, in main
    return args_func(args, parser)
  File "/Users/travis/Dropbox/work/github/conda/conda/conda/cli/main.py", line 207, in args_func
    args.func(args, p)
  File "/Users/travis/Dropbox/work/github/conda/conda-env/conda_env/cli/main_create.py", line 95, in execute
    installer.install(prefix, specs, args, env)
AttributeError: 'module' object has no attribute 'install'
```

Digging in, but it doesn't appear that `__import__` gives you what you expected @asmeurer.

PR #47, opened 2015-01-26
Fix get_dirname() as suggested on the Anaconda Support list

asmeurer:
See https://groups.google.com/a/continuum.io/d/msg/anaconda/7i5nL1cD5Zc/2nYvIJ2xRh0J

Note this is untested.

### DIFF
--- a/bin/env-activate
+++ b/bin/env-activate
@@ -33,7 +33,7 @@ _THIS_DIR=$(dirname "$_SCRIPT_LOCATION")
 
 # Load common functions
 get_dirname() {
-    echo "$(cd "$(dirname "$1")" && pwd)"
+    echo "$(dirname "$1")"
 }
 source "$(get_dirname ${_ABS_SCRIPT_LOCATION})/_conda-functions"
 


bin/env-activate
```diff
@@ -33,7 +33,7 @@ _THIS_DIR=$(dirname "$_SCRIPT_LOCATION")
 
 # Load common functions
 get_dirname() {
-    echo "$(cd "$(dirname "$1")" && pwd)"
+    echo "$(dirname "$1")"
```
asmeurer: Shellcheck recommends deleting the `echo` and just using `dirname "$1"`.

tswicegood:
Let me prefix this by saying my availability is very spotty this week - sorry, family issue.

------

This code was there for a reason, though it predates me.  The original version was [committed here][1] and references [this StackOverflow answer][2] which also reference show to handle situations where the target directory might not exist.

> **What if dir doesn't exist?**
> 
> As given above the function will fail and print on stderr if the directory path given does not exist. This may not be what you want. You can expand the function to handle that situation:
> 
> ```bash 
> #!/bin/bash
> get_abs_filename() {
>   # $1 : relative filename
>   if [ -d "$(dirname "$1")" ]; then
>    echo "$(cd "$(dirname "$1")" && pwd)/$(basename "$1")"
>   fi
> }
> ```

From the error message in the message, it would appear that this install somehow is missing `${ROOT_PREFIX}/bin/_conda-functions`.  I don't believe that's possible.

I do believe there's a clue in the way the error message was emailed:

> /Users/stevend2/anaconda/bin/env-deactivate:source:37: no such file or directory: ^[]7;file:///Users/stevend2/anaconda**/bin^G/Users**/stevend2/anaconda/bin/_conda-functions

Emphasis mine.  For some reason, `_ABS_SCRIPT_LOCATION` is not being set properly and includes that control character.  There's code in there to handle symlink resolution when setting that, and unfortunately I [did a horrible job documenting][3] how I came by that particular solution.  Apologies.

A quick search of Stack Overflow brought up some similar solutions.  I believe that's where that comes from, though I can't specifically remember what or where.

I'll help as I can this week -- let's keep this PR open as the place for communication.  Unfortunately, however, I'm going to be traveling this afternoon/evening and have spotty connections the rest of the week.

[1]: https://github.com/conda/conda/commit/72997d3b74380b42e72d983dace2301cb3e48a71
[2]: http://stackoverflow.com/a/21188136/161801
[3]: https://github.com/conda/conda/commit/1ea47a21ec1fd8009fdc268495b9fd0797b3ec44

tswicegood:
Just a quick bit of background: this code in question is attempting to resolve the full location of the root Anaconda installation.  It's entirely reasonable for this code to also be modified to simply do:

    _ABS_SCRIPT_LOCATION="$(conda info --root)/bin"

I believe I added `--root` *after* jumping through all of these hoops.

At any rate, what's needed here is not `{{ env }}/bin`, but `{{ conda root }}/bin` so I can load the `_conda-functions` file without having to ensure it was linked properly (it wouldn't be on existing environments).

Another possible solution would be to simply duplicate that code.  Originally, `_conda-functions` had more in it, but it's hard to share functions when some of those functions are necessary to figure out where to load the files from.

asmeurer:
Yeah, I think I remember making this change now.

asmeurer:
Be sure to run the conda activate tests before merging any changes here.

PR #52, opened 2015-02-03
Remove _conda-functions file (v2.0.2)

tswicegood:
Per a423ad228c7f73366279da95a8899dd27e4799d1.

> This violates DRY principles, but is introducting issues related
> determining the actual location of the conda root to load the functions
> file.  As such, it's easier to suffer a little duplication for the time
> being instead of dealing with bugs.


tswicegood:
Holding my nose and merging this even though binstar builds are failing.  We'll build them locally. :-/

PR #60, opened 2015-02-20
Fix windows activate.bat script

### LINKED ISSUE #59: activate.bat totally doesn't work with paths
```
>conda create -p D:\myenv python=2.7
...
Proceed ([y]/n)? y
...
#
# To activate this environment, use:
# > activate D:\myenv
#

>activate D:\myenv
No environment named "D:\myenv" exists in C:\Anaconda\envs
```
[In activate.bat](https://github.com/conda/conda-env/blob/616e3ddc6380f781e4dd586a0499676fcf0a8cf3/bin/activate.bat#L23):
```batch
if exist "%ANACONDA_ENVS%\%CONDA_NEW_ENV%\conda-meta" goto skipmissingenv
```

remram44:
Correctly handles paths (not names) arguments in activate.bat.

Fixes #59

Note: `CONDA_DEFAULT_ENV` is now always a full path.

### DIFF
--- a/bin/activate.bat
+++ b/bin/activate.bat
@@ -32,27 +32,27 @@ if "%CONDA_DEFAULT_ENV%" == "" goto skipdeactivate
     echo Deactivating environment "%CONDA_DEFAULT_ENV%"...
 
     REM Run any deactivate scripts
-    if not exist "%ANACONDA_ENVS%\%CONDA_DEFAULT_ENV%\etc\conda\deactivate.d" goto nodeactivate
-        pushd "%ANACONDA_ENVS%\%CONDA_DEFAULT_ENV%\etc\conda\deactivate.d"
+    if not exist "%CONDA_DEFAULT_ENV%\etc\conda\deactivate.d" goto nodeactivate
+        pushd "%CONDA_DEFAULT_ENV%\etc\conda\deactivate.d"
         for %%g in (*.bat) do call "%%g"
         popd
     :nodeactivate
 
-    set CONDACTIVATE_PATH=%ANACONDA_ENVS%\%CONDA_DEFAULT_ENV%;%ANACONDA_ENVS%\%CONDA_DEFAULT_ENV%\Scripts;
+    set CONDACTIVATE_PATH=%CONDA_DEFAULT_ENV%;%CONDA_DEFAULT_ENV%\Scripts;
     call set PATH=%%PATH:%CONDACTIVATE_PATH%=%%
     set CONDA_DEFAULT_ENV=
     set CONDACTIVATE_PATH=
 :skipdeactivate
 
-set CONDA_DEFAULT_ENV=%CONDA_NEW_ENV%
+set CONDA_DEFAULT_ENV=%ANACONDA_ENVS%\%CONDA_NEW_ENV%
 set CONDA_NEW_ENV=
 echo Activating environment "%CONDA_DEFAULT_ENV%"...
-set PATH=%ANACONDA_ENVS%\%CONDA_DEFAULT_ENV%;%ANACONDA_ENVS%\%CONDA_DEFAULT_ENV%\Scripts;%PATH%
+set PATH=%CONDA_DEFAULT_ENV%;%CONDA_DEFAULT_ENV%\Scripts;%PATH%
 set PROMPT=[%CONDA_DEFAULT_ENV%] $P$G
 
 REM Run any activate scripts
-if not exist "%ANACONDA_ENVS%\%CONDA_DEFAULT_ENV%\etc\conda\activate.d" goto noactivate
-    pushd "%ANACONDA_ENVS%\%CONDA_DEFAULT_ENV%\etc\conda\activate.d"
+if not exist "%CONDA_DEFAULT_ENV%\etc\conda\activate.d" goto noactivate
+    pushd "%CONDA_DEFAULT_ENV%\etc\conda\activate.d"
     for %%g in (*.bat) do call "%%g"
     popd
 :noactivate
--- a/bin/activate.bat
+++ b/bin/activate.bat
@@ -21,11 +21,17 @@ if not "%CONDA_NEW_NAME%" == "" goto skipmissingarg
 :skipmissingarg
 
 if exist "%ANACONDA_ENVS%\%CONDA_NEW_NAME%\conda-meta" goto usenamedenv
-    echo No environment named "%CONDA_NEW_NAME%" exists in %ANACONDA_ENVS%
-    set CONDA_NEW_NAME=
-    exit /b 1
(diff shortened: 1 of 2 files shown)


remram44:
Note that this currently tries names first and then paths, so if an environment "test" is in the current directory, it will only be tried if `$ANACONDA_ENVS/test` doesn't exist.

bin/activate.bat
```diff
@@ -32,27 +40,28 @@ if "%CONDA_DEFAULT_ENV%" == "" goto skipdeactivate
...
-    set CONDACTIVATE_PATH=%ANACONDA_ENVS%\%CONDA_DEFAULT_ENV%;%ANACONDA_ENVS%\%CONDA_DEFAULT_ENV%\Scripts;
+    set CONDACTIVATE_PATH=%CONDA_DEFAULT_ENV%;%CONDA_DEFAULT_ENV%\Scripts;
     call set PATH=%%PATH:%CONDACTIVATE_PATH%=%%
     set CONDA_DEFAULT_ENV=
     set CONDACTIVATE_PATH=
 :skipdeactivate
 
-set CONDA_DEFAULT_ENV=%CONDA_NEW_ENV%
-set CONDA_NEW_ENV=
+set CONDA_DEFAULT_ENV=%CONDA_NEW_PATH%
```
tswicegood: I'm just doing a quick read-thru of this code.  The way I read it, it's going to change the prompt in the environment to always be a path and not keep the name around.  Is that correct?  If that's the case, we need to modify this so it'll display just the environment name except when it's a full path that's been activated.  Can you take a stab at adding that behavior in?

tswicegood:
@remram44 this looks ok though I haven't tried it out in a VM yet.  I don't have a VM setup with Conda in it right now to test.  I do have that one question about what gets added to the prompt.  It's entirely possible I just misread the code though.  I always have to brush up on my Batch-fu when I start working with these files.  :-)

Thanks for putting this together.  I'll try to get everything setup to run this early next week.

bin/activate.bat
remram44: No, the prompt gets only the name (see line 58: `set PROMPT=[%CONDA_NEW_NAME%] $P$G`); it's the reason I have separate `CONDA_NEW_NAME` and `CONDA_NEW_PATH` variables. The name is either the venv name (if you entered a name) or the last component of the path (if you entered a path, see line 34, `CONDA_NEW_NAME=%%~ni`).

But the `CONDA_DEFAULT_ENV` gets the full path; I adapted deactivate.bat to expect that.

remram44:
No worries! This is not blocking for me by any means.

remram44:
@tswicegood: did you get a chance to review this?

tswicegood:
@remram44 I didn't get a chance to look at it before I took off for a conference the end of last week.  I've got quite a bit of Windows work ahead of me in the next few days, so I'll be taking a look at at this and getting it tested.  I think we'll be good for a merge and it'll be released as soon as it's merged.

Thanks for your patience and help on this!

gdementen:
Could we also keep the name of the "current" environment in another variable (when available)? I have a bunch of scripts which make use of it and extracting the name from the path seems backward to me.

remram44:
I don't remember what this is about.

If the conda team still has an interest I'll look into the change you need.

remram44:
Hello again. Is somebody interested in this at all? Is conda not maintained on Windows anymore?

bin/activate.bat
```diff
@@ -4,14 +4,14 @@ for /f "delims=" %%i in ("%~dp0..\envs") do (
     set ANACONDA_ENVS=%%~fi
```
wulmer: This PR seems to fix the problem with the activation of environments identified by an absolute path.
As all other environments are searched under `%ANACONDA_ENVS%` there remain the following problems:
* Activation of an environment in one of the directories configured as `envs_dirs` in the `.condarc` file is not possible (at least not by name, just by full path).
* Activation of an environment specified by a relative path is not possible. This is very handy if you have a virtual environment in all of your working directories. There, the activation is done by a simple `activate .\venv` instead of `activate %CD%\venv` (which are three! more characters to type).

The second point is not so important, but should be addressed also, but the first point really is essential. And to solve it, this requires that the `activate.bat` script knows about the conda environment directories. In my old pull request, this had been solved by creating a special hidden command in conda (`..checkenv`, kudos to @ognenduz) which does all the environment checking and of course knows about all the envs directories and all environments in them.
This functionality should now go into `conda-env`, something like `conda env list` but with output for reuse in the `activate.bat` script.

remram44: I agree that this PR doesn't cover all the features that should be there, it does however fix the ones that are there (and essential, honestly).

Of course I'm not going to do any more work until I'M CONFIDENT SOMEONE WILL REVIEW MY SHIT AT ALL. COME ON. THIS IS JUST RUDE NOW.

yoavram:
:+1: please merge this to conda!

msarahan:
I am sorry this got lost.  I am now responsible for Windows issues with Conda and Anaconda, and will give this a proper review by 10/9/15.

remram44:
That is good news. Don't worry, no one is holding their breath over this (since they are definitely dead).

yoavram:
@msarahan sounds good.
@remram44 thanks for your PR. I copied your `activate.bat` and put it in my scripts folder as `conda-activate.bat`. I use it to activate conda environments by folder (like I usually do with virtualenv) and it work great (so another positive tester, using conda 3.18, root python 3.4, Win 7-64bit).

msarahan:
Thanks for your PR.  I have updated it for current PATH (we add Library/bin also now).  I also enclosed all paths with quotes for space safety.  Sorry for the long wait, but thank you for your patience.

yoavram:
One comment: on my windows shell (ConEmu with the Cmdr configuration), once I activate a conda env with this bat script, I don't see the name of the git branch on the prompt anymore - but I do are the name of the env.

remram44:
I kept this behavior from before, this is in no way affected by my patch.

The code goes like this:

```batch
set PROMPT=[%CONDA_NEW_NAME%] $P$G
```

It totally ignores the fact that your previous prompt might not be `$P$G`. I remember thinking about changing it to:

```batch
set PROMPT=[%CONDA_NEW_NAME%] %PROMPT%
```

however there is a risk of appending more and more stuff to the prompt as you switch environments... Storing the "initial prompt" on the first activation is also an option, but it plays badly with other things you can activate... do we care?

msarahan:
I think we do care, but it is a question of how high priority this is. It
seems to me that Remi's code provides a good example of stripping strings
out of PATH, and something similar could prevent appending more and more
stuff to PROMPT.

I don't presently have time to implement THIS myself, but I'm happy to
review any PR.

yoavram:
I'll open a separate issue for this, then.

PR #64, opened 2015-03-20
Allow third party integration

tswicegood:
This came out of a conversation with @quasiben about ways to avoid adding a ton of duplicate commands when dealing with add-ons such as the tooling for Anaconda Cluster.  Currently, the command must look something like this:

``` console
$ conda cluster --cluster-name=foo create ... rest of arguments
# or...
$ conda cluster foo create ... rest of arguments
```

This new code allows integration directly with `conda env create` by allowing that package to register its own parameters via `conda_env.cli.main_create.configure_parser` and a custom execution method with `conda_env.cli.main_create.execute`.  Under this system, the API would be:

``` console
$ conda env create --cluster foo ... rest of arguments
```

The idea here is to allow hooking in to any command and circumventing it.  In the case of Anaconda Cluster, it's to take a command that runs locally and dispatch it to a cluster of machines.
## Tasks
- [x] Proof of concept
- [x] Build out tests for all decoration code
- [x] Allow code to act as full middleware where it can change behavior before _and_ after the main execute() has run
- [x] Document new extension point

### DIFF
--- a/conda_env/cli/main_create.py
+++ b/conda_env/cli/main_create.py
@@ -13,11 +13,9 @@ from conda.misc import touch_nonadmin
 from ..env import from_file
 from ..installers.base import get_installer, InvalidInstaller
 from .. import exceptions
+from . import helpers
 
-ENTRY_POINTS = {
-    "configure_parser": "%s.configure_parser" % __name__,
-    "execute": "%s.execute" % __name__,
-}
+ENTRY_POINTS = helpers.generate_entry_points(__name__)
 
 description = """
 Create an environment based on an environment file
--- a/conda_env/cli/main_create.py
+++ b/conda_env/cli/main_create.py
@@ -67,12 +67,6 @@ def configure_parser(sub_parsers):
     )
     common.add_parser_json(p)
 
-    entry_points = pkg_resources.iter_entry_points(
-        ENTRY_POINTS["configure_parser"]
-    )
-    for entry_point in entry_points:
-        entry_point.load()(p)
-
     p.set_defaults(func=execute)
 
 
--- a/conda_env/cli/main_create.py
+++ b/conda_env/cli/main_create.py
@@ -30,14 +30,21 @@ examples:
 """
 
 
-def sub_parser(p):
-    print("sub_parser invoked")
-    return sub_parser.next(p)
+def say_configure_parser(sub_parsers):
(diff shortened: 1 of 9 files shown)


tswicegood:
@quasiben, this is the implementation we talked about earlier.  I do want to modify this code so the hook can allow execution of the before and after, but for your purposes I think this will work.

tswicegood:
@quasiben, I've added docs to this.  Let me know if you run in to any trouble or if anything needs clarification.

tswicegood:
@quasiben were you able to give this a go?

tswicegood:
@pzwang @ilanschnell @srossross: this is the PR I mentioned in the meeting this morning.  Reading through the documentation shows a running example that I created with this.  The whole idea here is to allow external packages to grab ahold of the execution of the commands and:

* control the parameters that are configured for the user (normally by adding additional parameters)
* add additional behavior before or after the main execution script is run

This is definitely a first pass.  I would like to enhance the command execution to a class-based system so that users have finer grained controls over how a given command executes.  An excellent example is the middleware in the documentation not being able to determine the environment name because the name information is locked up inside `execute`.

conda_env/cli/main_create.py
```diff
@@ -25,6 +30,24 @@
 """
 
 
+def say_configure_parser(sub_parsers):
```
tswicegood: This code needs to be removed prior to merge / release.

setup.py
```diff
@@ -40,12 +40,20 @@
...
     ],
     description="tools for interacting with conda environments",
-    long_description=open('README.rst').read(),
+    long_description=open('README.rst', 'rb').read().decode('UTF-8'),
     packages=[
         'conda_env',
         'conda_env.cli',
         'conda_env.installers',
     ],
+    entry_points={
```
tswicegood: This code needs to be removed prior to merge / release.

tswicegood:
FWIW, thinking about the idea of a `Command` object sparked some ideas.  I've coded those up in [refactor/move-to-command-objects][1] and modified the `main_create` module to use the new code.  Would be curious for some thoughts on that.  Of particular note is how that exposes a `command.name` to the `say_execute` extension now so the extension can reach into the command and pull out information that was originally locked inside a function.

[1]: https://github.com/conda/conda-env/compare/feature/third-party-integration...refactor/move-to-command-objects

PR #69, opened 2015-04-08
avoid enlarging prompt with multiple activate

gdementen:
If you activate several environments in turn on Windows, your prompt gets ever larger. eg, by activating "edge2" then "edge3" then "liam2" then "edge3", I just got:

[edge3] [liam2] [edge3] [edge2] C:\Users\gdm\devel\liam2>

I guess what you intended was something like above where only the initial prompt is remembered and recalled.

### DIFF
--- a/bin/activate.bat
+++ b/bin/activate.bat
@@ -48,8 +48,11 @@ set CONDA_DEFAULT_ENV=%CONDA_NEW_ENV%
 set CONDA_NEW_ENV=
 echo Activating environment "%CONDA_DEFAULT_ENV%"...
 set PATH=%ANACONDA_ENVS%\%CONDA_DEFAULT_ENV%;%ANACONDA_ENVS%\%CONDA_DEFAULT_ENV%\Scripts;%PATH%
-set CONDA_OLD_PROMPT=%PROMPT%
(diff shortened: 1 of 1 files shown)


tswicegood:
Whoa, good catch.  That's definitely not the behavior that should be there.  It's supposed to deactivate the current environment, then activate the new one.

gdementen:
Could you fix this bug one way or another? It severely impacts the usability of conda on Windows...

tswicegood:
@ilanschnell @asmeurer can one of you take a look at this?

kerrywatson1:
@ilanschnell @asmeurer this will affect the way we write documentation for Windows users -- we currently have to tell them to deactivate an environment before they activate another.

malev:
It has been attacked with PR #118

gdementen:
Thanks for finally tackling this. My fix was better IMO though. If you activate twice and deactivate, you get:
"[env1] XXX" using the merged fix, which is incorrect (no env is actually activated at that point), while with my fix, you get the actual original prompt.

malev:
@gdementen agree, do you mind updating your PR and pointing it to develop so I can merge it? tnx

PR #73, opened 2015-04-23
Download and create evironments from binstar

malev:
# TODO
* [x] Add tests
* [x] Isolate Binstar client artifacts into a single module
* [x] What if the user is not logged into Binstar? Will show error message
* [x] ~~What if version is not in Binstar?~~ Leaving out versions for now
* [x] Allow other environment sources (github, web, etc) 
* [x] ~~Print what environment.yml version is being installed~~ Versions are out of the game for now
* [x] Create env without downloading environment.yaml
- [x] ~~Overwrite name with `-n`~~
- [x] Deprecate `--name`


conda_env/cli/main_create.py
```diff
@@ -22,6 +23,7 @@
     conda env create -n=foo
     conda env create -f=/path/to/environment.yml
     conda env create --name=foo --file=environment.yml
+    conda env create --name=foo --binstar=username/foo
```
malev: I agree with @tswicegood it would be great to have this as:
```
conda env create --name=wakari malev/wakari
# or
conda env create malev/wakari
```
But I'm not sure where to have this discussion

conda_env/utils/binstar.py
```diff
@@ -0,0 +1,61 @@
...
+from conda.cli import common
+from conda.resolve import normalized_version
+from binstar_client import errors
+from binstar_client.utils import get_binstar
+
+
+ENVIRONMENT_TYPE = 'env'
+
+
+def download_environment_file(package_spec, filename, json=False):
```
malev: I'm planning to split this function into a class with a bunch of methods

conda_env/cli/main_create.py
tswicegood: This syntax is incorrect -- using it we would have to add `--github` or `--gist` or `--whatever` for every single remote location we want to support.  There are two parts to this, adding the concept of an Environment Spec and then the backend to lookup those specs against known sources.  I'm going to break it out into an issue on its own, then reference this PR.

tswicegood: They should conform to an API like this (doesn't have to be this specifically, but this gives you an idea):

```python
class EnvironmentLoader(object):
    def can_handle(self, name):
        # return True if you can process, False if you can't

    def load(self, name):
        # return an Environment object
```

The main thing is that a processor needs to be able to say that it can deal with a request and then actually deal with the request.

So from there, imagine a Binstar one that might look like this (Github would look similar):

```python
class BinstarLoader(EnvironmentLoader):
    def can_handle(self, name):
        if "/" not in name:
            return False  # Bail early
        # attempt to load from binstar, return True if its found
    
    def load(self, name):
        # fetch environment.yml from binstar
        return env.Environment(url=binstar_url, **data_loaded_from_binstar)
```

This should be done by loading a list of backends (look at the [get_installer][1] for examples) that are possibly able process the backend.  This is a good case for `pkg_resources.iter_entry_points`.  Now that we install `pip` by default in every environment that `python` is installed there shouldn't be any resistance to using it (see [this diff][2] for examples for loading entry points).

[1]: https://github.com/conda/conda-env/blob/develop/conda_env/installers/base.py#L11-L15
[2]: https://github.com/conda/conda-env/pull/64/files#diff-a6b0af01b802da645b8d8a37879fa3c6R40

malev: Sounds great to me!

conda_env/exceptions.py
```diff
@@ -7,3 +7,38 @@ def __init__(self, filename, *args, **kwargs):
...
+
+class PackageNotFound(CondaEnvException):
+    def __init__(self, username, packagename, *args, **kwargs):
+        msg = '{}/{} not found'.format(username, packagename)
+        self.username = username
+        self.packagename = packagename
+        super(PackageNotFound, self).__init__(msg, *args, **kwargs)
+
+
+class LoaderNotFound(CondaEnvException):
```
tswicegood: In order for this to be caught by [`conda.cli.common.arg_func`][1] you need to also extend `RuntimeError`.  I'm fine with the inheritance being `LoaderNotFound` extends `CondaEnvRuntimeError` extends `CondaEnvException` and `RuntimeError`.  That provides a nice clean way to capture all `CondaEnvRuntimeError` exceptions.

[1]: https://github.com/conda/conda/blob/master/conda/cli/main.py#L209-L210

malev: done!

conda_env/cli/main_create.py
```diff
@@ -51,22 +53,36 @@ def configure_parser(sub_parsers):
         '-q', '--quiet',
         default=False,
     )
+    p.add_argument(
+        'handle',
```
tswicegood: Let's change this to name and deprecate the `--name` tag.

conda_env/cli/main_create.py
```diff
@@ -51,22 +53,36 @@ def configure_parser(sub_parsers):
...
-            conda env create.""").lstrip()))
-
-        common.error_and_exit(msg, json=args.json)
+
+    if args.handle:
+        loader = get_loader(args.handle)
+        if loader is not None:
+            env = from_yaml(loader.get(), args.name)
+        else:
+            raise exceptions.LoaderNotFound(args.handle)
```
tswicegood: What about letting `get_loader` throw this exception?  This would simplify this code block and make it easier to test.

conda_env/cli/main_create.py
malev: two problems:

1) If I run:
` conda env create malev/my-project` then I'll have an environment named `my-project` or whatever environment's name I have on my environment.yml. What I am saying is: I define `malev/my-project` as name and then the environment will be named differently. I think it looks confusing.
2) Deprecating `--name` should be in another PR. This change will affect multiple parts of the code and documentation

makes sense?

conda_env/cli/main_create.py
```diff
@@ -51,22 +53,36 @@ def configure_parser(sub_parsers):
...
+            env = from_file(args.file)
+        except exceptions.EnvironmentFileNotFound as e:
+            msg = 'Unable to locate environment file: %s\n\n' % e.filename
+            msg += "\n".join(textwrap.wrap(textwrap.dedent("""
+                Please verify that the above file is present and that you have
+                permission read the file's contents.  Note, you can specify the
+                file to use by explictly adding --file=/path/to/file when calling
+                conda env create.""").lstrip()))
+
+            common.error_and_exit(msg, json=args.json)
```
tswicegood: Unrelated, but if you wanted to do some code weeding, this should be an exception as well.

conda_env/env.py
```diff
@@ -46,6 +46,14 @@ def from_environment(name, prefix):
...
 
+def from_yaml(yamlstr, name=None):
+    """Load and return a ``Environment`` from a given ``yaml string``"""
+    data = yaml.load(yamlstr)
+    if name is not None:
+        data['name'] = name
+    return Environment(**data)
+
+
 def from_file(filename):
```
tswicegood: This function should be refactored to dispatch to `from_yaml` as well.

conda_env/loaders/__init__.py
```diff
@@ -0,0 +1,25 @@
+import importlib
+from ..exceptions import CondaEnvException
+ENTRY_POINT = 'conda_env.loaders'
+LOADERS = ['binstar']
+
+
+class InvalidLoader(Exception):
```
tswicegood: This belongs over in `conda_env.exceptions`

conda_env/loaders/binstar.py
```diff
@@ -0,0 +1,79 @@
...
+
+    def valid_handle(self):
+        return re.match("^(.+)/(.+)$", self.handle)
+
+    def package_exists(self):
+        """
+        Checks whether a package exists on binstar or not.
+        :return: True or False
+        """
+        username, packagename = self.parse()
```
tswicegood: Since these values are used in multiple places, I would parse it immediately and add them as properties.

malev: I was hesitant to do that because I didn't wanted to generate too much side effect on a validation. But make sense

tswicegood:
Everything looks good, minus lack of tests. :-)

One thing that did strike me, the [`BinstarLoader`][1] could be made to provide a little more nicety around the API.  It's basically [value object][2] in the way it's implemented, but some of the functionality (`parse` for example) feel a little off.  That actually makes a lot of sense to me to make those value objects that can return an environment.  For example, consider the following:

```python
loader = loader.get_loader(name)
env = env.from_yaml(loader.get())
```

Contrast this to this:

```python
env = get_loader(name).environment
```

I could see the above as a wrapper around essentially doing:

```python
env = None
for LoaderClass in get_all_loaders():
    loader = LoaderClass(name)
    if not loader.can_process:
        continue
    env = loader.environment
    break
```

Anyhow, that's the one thought I have.  I would make the loader work more like an object that represents a possible remote environment spec (maybe that's the name `BinstarSpec`, `FileSpec`, `GitHubSpec`) that is either valid or not.  I'll leave the name to you, but I do like the idea of a file that:

* Is an unchanging value object that represents a possibly valid remote environment specification
* Can answer whether it's valid or not (`is_valid`, `can_process`, something like that)
* And has an `environment` property that creates a valid `Environment` object when invoked on a valid loader/spec.

[1]: https://github.com/conda/conda-env/pull/73/files#diff-1bcad1450a3e3aac5cae72df7a1ce929R18
[2]: http://martinfowler.com/bliki/ValueObject.html

conda_env/cli/main_create.py
malev: I like it but in another PR. #76

conda_env/env.py
malev: I have no idea what are you talking about :lollipop:

conda_env/cli/main_create.py
tswicegood: I don't think issue 1 is an issue.  When you run `hub clone tswicegood/conda` you get `conda` as a sub-directory as the portion before the `/` is part of the namespace.  Same with `git clone git://github.com/tswicegood/conda.git`.  The last element in the URI provides the leaf node, which is what's assumed to be the name as that's where the tree terminates.

I'd prefer to do the name deprecation as part of this so we have it all in one take instead of multiple releases.  This is driven by wanting to do `conda env create web-dev gulp` and have that translate to `conda create -n web-dev gulp`.  I can see another case where `conda env create tswicegood/web-dev broccoli` would create an environment from `tswicegood/web-dev` but also add `broccoli` to it.

With the deprecation, I see `conda env create tswicegood/web-dev` and `conda env create --name tswicegood/web-dev` being equivalent with the latter displaying a deprecation warning.

conda_env/env.py
tswicegood:
```python
def from_file(filename):
    if not os.path.exists(filename):
        raise exceptions.EnvironmentFileNotFound(filename)
    with open(filename, "rb") as fp:
        return from_yaml(fp.read(), filename=filename)
```

Also, `from_yaml()` would need to be adjusted to take `kwargs`.  I think that works better than `name` as the only value you can provide explicitly.

malev: ok now I see what you are saying! You are talking about `from_file`!!

conda_env/cli/main_create.py
malev: ok, let's deprecate `--name` you are right about github.
I'm gonna focus on having:
```
conda env create --name malev/wakari
`--name` is being deprecated, just use:
    conda env create malev/wakari
```

tswicegood: :+1:

conda_env/cli/main_create.py
```diff
@@ -33,54 +33,63 @@ def configure_parser(sub_parsers):
...
+            print("`--name` is deprecated. Use:\n"
+                  "  conda env create {}".format(args.old_name))
+
+        for Spec in all_specs:
+            spec = Spec(name)
+            if spec.can_process():
+                env = from_yaml(spec.environment)
+                args.name = env.name
+
+    else:
```
tswicegood: The way this `if/else` is setup it changes the behavior subtly.  It used to be that `--name` was an override to anything that was found in the `--file` (implied or explicit), now you either use a name or a file, but both can't be used together.  I'm going to open up a PR against this with some tweaks.

conda_env/specs/__init__.py
```diff
@@ -0,0 +1,4 @@
+from binstar import BinstarSpec
```
tswicegood: FYI, this will get you in trouble.  It's a implied relative import that breaks in Python 3.  Though I'm setting a bad example in this code base, it's always a good idea to include `from __future__ import absolute_import`.  This needs to be `from .binstar import BinstarSpec`.

conda_env/specs/tests/test_binstar.py
```diff
@@ -0,0 +1,50 @@
...
+from ...exceptions import EnvironmentFileDoesNotExist
+
+
+class TestBinstarSpec(unittest.TestCase):
+    def test_invalid_handler(self):
+        spec = BinstarSpec('invalid')
+        self.assertEqual(spec.valid_handle(), False)
+        self.assertEqual(spec.can_process(), False)
+
+    def test_package_not_exist(self):
```
tswicegood: :+1: on the mock usage here.  Very nice.

conda_env/specs/binstar.py
```diff
@@ -0,0 +1,92 @@
...
+                raise EnvironmentFileDoesNotExist(self.handle)
+
+            versions = {normalized_version(d['version']): d['version'] for d in file_data}
+            latest_version = versions[max(versions)]
+            file_data = [data for data in self.package['files'] if data['version'] == latest_version]
+            req = self.binstar.download(self.username, self.packagename, latest_version, file_data[0]['basename'])
+            if req is None:
+                raise EnvironmentFileNotDownloaded(self.username, self.packagename)
+            self._environment = req.raw.read()
+        return self._environment
```
tswicegood: What I was thinking here was that this would return the full `Environment` object, not just the yaml.

tswicegood:
Two general notes:

* Tests should be over in the `./tests/` directory so they're not packaged up with the script.
* There's no tests around the new exceptions.

I've got an idea I want to experiment a little bit with in code -- I'll open up a PR against it shortly.

conda_env/specs/yaml_file.py
```diff
@@ -0,0 +1,19 @@
...
+from .. import env
+
+
+class YamlFileSpec(object):
+    def __init__(self, filename=None, name=None, **kwargs):
+        self.filename = filename
+        self.name = name
+
+    def can_handle(self):
+        self._environment = env.from_file(self.filename)
```
malev: If `self.filename` does not exist we are going to have an exception here. The exception is being captured in `main_create` module and we are showing a file not exist error. That's fine, but `args.filename` is not mandatory and it shouldn't because we can create an environment from a remote location. Looks like YamlFileSpec should be the last spec to execute. If that's the case we will need to ensure `all_specs` order and that seems to be strange. Makes sense?

malev: I have an idea about this.
Normally a `detect` method should raise an exception when failing to detect. Following that behavior we can raise an exception that has logged why every spec replied `False` to `can_handle`. And we can show that error to the user. Something like:
```
Couldn't find an environment.yml file.
Couldn't find malev/project in Binstar.org
...
```

tswicegood: The problem throwing from inside `can_handle` is that it provides a name that appears to answer a yes/no question.  Thinking in terms of true or false, if I ask if you want to go to lunch I expect you to say yes or no, not yes or flip a table. :-)

Exceptions are for behavior that's exceptional -- a behavior you wouldn't expect.  It's reasonable for `detect` to fail with an exception as you expect `detect` to work, but `can_handle` implies a true/false.

Given your example, it might even be worth adding an `obj.error_message` to each loader so it explains what it couldn't do.  That way:

```python
if not loader.can_handle:
    errors.append(loader.error_message)
    continue
env = loader.detect()
break
```

Then detect for `env is None` and display the error messages if that's the case, otherwise proceed.

malev:
@tswicegood I think this is good to go

tswicegood:
Checking into this now.

tests/specs/test_yaml_file.py
```diff
@@ -0,0 +1,24 @@
...
+        spec = YamlFileSpec(name=None, filename='not-a-file')
+        self.assertEqual(spec.can_handle(), False)
+
+    def test_environment_file_exist(self):
+        with mock.patch.object(env, 'from_file', return_value={}):
+            spec = YamlFileSpec(name=None, filename='environment.yaml')
+            self.assertTrue(spec.can_handle())
+
+    def test_get_environment(self):
+        with mock.patch.object(env, 'from_file', return_value={}):
```
tswicegood: I love being able to read 3 - 5 line test methods when doing a review.  Really helps me get a handle on what's going on.  Nice work!

This is me showing my TDD habits so feel free to take with a grain of salt.  A good habit to get into with assertions like this is to test with some sort of random data.  For this particular function, there are two important things:

* `from_file` is given the `filename` that's provided
* whatever is returned from `from_file` is returned to the user

Can you add a test that verifies that is provided to the patched `from_file` and ensure that the value passed in as `filename` is passed along?  Having a random filename can help you verify that whatever is provided to `YamlFileSpec` is what actually gets passed in to `from_file`.

For the second test, I could write out `YamlFileSpec` to do something like this and this test would pass since `{}` is hard coded:

```python
class YamlFileSpec(object):
    def __init__(self, **kwargs):
        self.environment = {}
```

Adding a `random_return = random.randint(100, 200)` and using that value for `return_value` of the mock object means that it verifies that the randomly generated value that the mock object returns is what actually got returned.

---

Quick digression: you have to be very careful to find the right balance with mock objects.  Too many can be a sign of something going wrong (too complex, too many dependencies, etc).  You can also end up with tests that are too isolated from the rest of the system so the tests still pass but the code doesn't work like it's expected to.

`from_file` [is tested][1] (note to self, I should have written more) so given a valid `filename` you know that it will return what you're looking for.  In that case, you don't need to test `from_file`, so mocking it and making some assumptions about it is the sweet spot for mocks.

[1]: https://github.com/conda/conda-env/blob/develop/tests/test_env.py#L31-L44

conda_env/specs/yaml_file.py
```diff
@@ -0,0 +1,25 @@
+from .. import env
+from ..exceptions import EnvironmentFileNotFound
+
+
+class YamlFileSpec(object):
+    _environment = None
+
+    def __init__(self, filename=None, name=None, **kwargs):
+        self.filename = filename
+        self.name = name
```
tswicegood: Is there a reason to store `name` here?  Since this takes `**kwargs`, we could just leave it in there and only pull out the relevant arguments.

conda_env/specs/binstar.py
```diff
@@ -0,0 +1,111 @@
...
+    @property
+    def environment(self):
+        """
+        :raises: EnvironmentFileNotDownloaded
+        """
+        if self._environment is None:
+            versions = {normalized_version(d['version']): d['version'] for d in self.file_data}
+            latest_version = versions[max(versions)]
+            file_data = [data for data in self.package['files'] if data['version'] == latest_version]
+            req = self.binstar.download(self.username, self.packagename, latest_version, file_data[0]['basename'])
```
tswicegood: Is there not a way to download the latest package for a given package?  That seems like logic that should be on the server side, not the client.

conda_env/specs/binstar.py
```diff
@@ -0,0 +1,111 @@
...
+try:
+    from binstar_client import errors
+    from binstar_client.utils import get_binstar
+except ImportError:
+    raise CondaEnvException("Binstar not installed")
+
+ENVIRONMENT_TYPE = 'env'
+
+
+class BinstarSpec(object):
```
tswicegood: Not a hard requirement, but there's a lot in this object that I would refactor away into `conda_env.utils.binstar`.  All of the code that interacts with the results of binstar information should really be in the client library.  I should be able to do:

```python
pkg = binstar.Package("darth/deathstar")
if not pkg.has_environments:
    # handle error state
else:
    return env.from_yaml(pkg.environments.latest.raw.read())
```

If we can isolate the code into `conda_env.utils.binstar`, then we can adjust to that new API if/when it's available without having to touch this code.  It also makes it easier to pull that code into the binstar API if we want to just move it over there.

If you don't have the bandwidth right now for the change, do you mind putting in a `TODO`?

conda_env/specs/binstar.py
```diff
@@ -0,0 +1,111 @@
...
+    """
+
+    _environment = None
+    _username = None
+    _packagename = None
+    _package = None
+    _file_data = None
+    msg = None
+
+    def __init__(self, filename=None, name=None, **kwargs):
```
tswicegood: Like `YamlFileSpec` below not needing `name`, I think this one can drop `filename`.

tswicegood:
Done with the review -- just a couple of comments.  The only big thing I see here is the refactoring of `BinstarSpec`.  I'm ok with that just being a `TODO` for now and coming back to it later.

I do want to run the code and play with it a bit, but the review looks good.  Nice work, señor!

FYI, I should be in the office Monday -- let's plan on getting it merged and tagged then so I can walk ya through what I do.

conda_env/specs/binstar.py
malev: should be on the client but it's not :cry: I guess we can add it later

conda_env/specs/binstar.py
malev: https://github.com/conda/conda-env/commit/9a9c2f161d92e164efcdce44fba8ef04a2987def

conda_env/specs/binstar.py
malev: I believe we need name here. https://github.com/conda/conda-env/blob/env-create-binstar/conda_env/specs/binstar.py#L53

malev:
I fixed most of the comments. We can talk about his anytime

tswicegood:
Looks good to me! :+1:

Issue #63, opened 2015-03-20
activate should export an environment variable with the full path to the environment


asmeurer:
I've had a couple people request this. `CONDA_DEFAULT_ENV` is not what they want because they want a full path.

asmeurer:
Alternately, we should add `conda info --default-env`. Parsing `conda info --json` is not something people should be doing from the shell.

tswicegood:
Hmm.  I can see both options as a good way to do this.  Having an environment variable is nice as it's available all sub-processes as well, though.  What to call it is the question.  At first pass, `CONDA_CURRENT_PREFIX` comes to mind.

I'd prefer not to use `DEFAULT` in the name because no matter how technically correct, it's confusing.  There's certain baggage that comes along with that word.

bkreider:
I have a good use case for this.  The openssl commands i put into FD last week.  I'd like to do something like this for non-python binaries that have data in the $PREFIX:
```
openssl s_client -connect store.continuum.io:443 -CAfile $PREFIX/ssl/cacert.pem
```

tdhopper:
@tswicegood: `default` would be consistent with the existing `CONDA_DEFAULT_ENV` and `conda info --default-prefix`.

tswicegood:
@tdhopper yeah -- what this should be is something more like `VIRTUAL_ENV`.  We might even simply do `CONDA_ENV`.  Actually, thinking along those lines, how about `CONDA_ENV_PATH`?

---

Regarding the use of default, I hope to start phasing it out.  I've yet to meet any developer who wasn't confused by it because the term has so much implicit meaning.  If you know the technical implementation of Conda, it starts to make sense, but users shouldn't have to understand every detail about Conda in order to be proficient and make mental models about what it is doing.

Not to worry, though.  No plans on removing it any time soon.

tdhopper:
I like `CONDA_ENV_PATH`.

malev:
We already have `CONDA_ENV_PATH` working everywhere but windows.
There is a patch coming for windows #139

ijstokes:
I'm worried there is something broken with this functionality.  Notice how `$CONDA_ENV_PATH` is totally confused through the sequence of operations I do below:

```bash
(sparkkernel)ijstokes@petawawa sbt-deps $ conda --version
conda 3.15.1
(sparkkernel)ijstokes@petawawa sbt-deps $ env | grep -i conda
[...]

PR #80, opened 2015-05-08
Set CONDA_ENV_PATH environmental variable to active env root path

tdhopper:
As discussed in [this issue](https://github.com/conda/conda-env/issues/63)

### DIFF
--- a/bin/activate
+++ b/bin/activate
@@ -67,6 +67,8 @@ if (( $? == 0 )); then
         export CONDA_DEFAULT_ENV="$@"
     fi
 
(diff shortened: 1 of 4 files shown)


tswicegood:
Sorry for the delay -- this looks good to me.  Merging and will release with the next release.  Thanks for the contribution! :+1:

tacaswell:
Thanks :beers:

tswicegood:
Unless someone really needs this, I'm going to not release it until Monday.  We've got PR #73 which should be merged then and I'll release it then.

tacaswell:
I do not need it over the weekend (and if I did I would just build my own package).

PR #84, opened 2015-06-01
Add support for Dependencies.add with dicts

campos-ddc:


malev:
What would be the usage of this?

Issue #83, opened 2015-05-27
Feature request: recursively process other environment.yml files, and add support for setting environment variables


campos-ddc:
Hi, first of all, thank you for your work in conda/conda-env, these tools are looking great!

Small preface, this feature request might have some relation with #65 and #21.

# Problem
In the company I work for (https://github.com/ESSS), we have the following development process:

* We have many external libraries dependencies, installing them via `conda` works just fine, since they are packages available in our channel (such as numpy, pytest, etc)
* We have many internal libraries that are always in 'development' mode, for these, we do not want to create packages during development. We commonly work in multiple libraries at once, and it would greatly hinder our workflow if we had to recreate packages whenever we make a small change in a dependency and want to test it in another library.

# Scenario
Say for example that we have 2 internal libraries, `DEV-A` and `DEV-B`. `DEV-A` depends on the external library `EXT-A`, while `DEV-B` depends on external libraries `EXT-B` and `EXT-C`.

In this scenario, when working in `DEV-B` I want to be able to:

* 'conda env' a single yaml file for `DEV-B`, that will recursively process the `DEV-A` yaml
* Avoid duplication of dependencies. Since `DEV-B` already depends on `DEV-A`, I should not have to re-declare dependencies that are inherited from `DEV-A`
* Set environment variables necessary to run `DEV-B` in development mode. Since this library is not being installed in a conda virtual environment, I need a way to define paths for its source code and shared libraries.
* Set aliases to be enabled in this environment.

# Proposed solution

I imagined a file that would look something like this, with a few options regarding how to define the environment variables:

```yaml
name: DEV-B

dependencies:
- EXT-2==2.4
- EXT-3==0.3.3

# Recursively process these other environment.yaml files
recurse:
- {{ root }}/../DEV-A/environment.yaml

# Environment option 1 -----------------------------------------------------------------------------
# Add these environment variables. Lists will be joined with os.pathsep. {{ root }} is replaced by
# jinja2, and points to the directory containing this file.
environment:
- PATH:
  - {{ root }}/tools/
  - {{ root }}/artifacts/libs [win]
  
- LD_LIBRARY_PATH:
  - {{ root }}/artifacts/libs [unix]

- PYTHONPATH:
  - {{ root }}/source/python
  - {{ root }}/artifacts/libs

- INCLUDE:
  - {{ root }}/source/c++
  - {{ root }}/source/bindings

alias:
- ptrun: python {{ root }}/tools/ptrun.py

[...]

PR #66, opened 2015-04-01
ENH : sort activate.d scripts to control order

tacaswell:
Sort the found shell scripts to allow the user to control which order
the scripts are run in.

I don't know enough to make the parallel change in the windows files.

The reason I want this is because I am trying to use conda packages to manage the setup/teardown of a test environment where I need to start a server-like thing, but some environmental variables need to be set first.

It currently sorts on the path to the scripts which is fine if you don't have anything in directories.  I am not convinced that this is not the desired behavior, just aware of the concern.

PR #85, opened 2015-06-01
Add support for activate/deactivate scripts

campos-ddc:
Scripts listed under 'activate' and 'deactivate' contexts in environment.yml will be linked into $PREFIX/etc/conda/(de)?activate.d, where they will be executed by 'source (de)?activate'

Relates to #83

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -39,8 +39,9 @@ All of the usage is documented via the ``--help`` flag.
 -------------------
 conda-env allows creating environments using the ``environment.yml``
 specification file.  This allows you to specify a name, channels to use when
-creating the environment, and the dependencies.  For example, to create an
-environment named ``stats`` with numpy and pandas create an ``environment.yml``
+creating the environment, dependencies and activate/deactivate scripts.
+
+To create an environment named ``stats`` with numpy and pandas create an ``environment.yml``
 file with this as the contents:
 
 .. code-block:: yaml
@@ -60,7 +61,7 @@ Then run this from the command line:
     [      COMPLETE      ] |#################################################| 100%
     #
     # To activate this environment, use:
-    # $ source activate numpy
+    # $ source activate stats
     #
     # To deactivate this environment, use:
     # $ source deactivate
@@ -73,3 +74,42 @@ You can override the name of the created channel by providing either ``-n`` or
 ``--name`` and a valid environment name.  Likewise, you can explicitly provide
 an environment spec file using ``-f`` or ``--file`` and the name of the file you
 would like to use.
+
+``environment.yml`` examples
+----------------------------
+
+Name and dependencies
+^^^^^^^^^^^^^^^^^^^^^
+
+.. code-block:: yaml
+
+    name: stats
+    dependencies:
+      - numpy
+      - pandas
+
+Name and version specific dependencies
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+.. code-block:: yaml
+
+    name: stats
+    dependencies:
+      - numpy==1.8
+      - pandas==0.16.1
+
+
+Activate/deactivate scripts
+^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+.. code-block:: yaml
+
+    name: oracle
+    dependencies:
+      - oracle_instantclient
+    
+    # Note that relative paths are relative to this file's location
+    activate:
+      - set_oraclehome.sh
+    deactivate:
+      - unset_oraclehome.sh
--- a/conda_env/cli/main_create.py
+++ b/conda_env/cli/main_create.py
@@ -105,6 +105,25 @@ def execute(args, parser):
             )
             return -1
 
+
+    # TODO should _link be moved to conda.misc?
+    # Use soft links to allow 'activate' scripts to obtain paths relative to its location
+    from conda.install import _link, LINK_SOFT
+
+    # Link activate scripts
+    if env.activate:
+        activate_dir = os.path.join(prefix, 'etc', 'conda', 'activate.d')
+        if not os.path.isdir(activate_dir): os.makedirs(activate_dir)
+        for i, activate_script in enumerate(env.activate):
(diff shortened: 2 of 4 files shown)


tacaswell:
:+1: 

This is independent of the activate/deactivate scripts that are install per-package correct?

campos-ddc:
Yes, unless you try to install a script with the same name as one obtained
from a package.

Note that this problem already exists if two packages have files with the
same name.

campos-ddc:
Oops, I actually rename files to 0.sh, 1.sh etc..

So there would only be an issue if a package contains a file with those names.

tacaswell:
https://github.com/conda/conda-env/pull/66 is vaguely relevant to this and stalled by my lack of windows knowledge. 

I am not sure renaming is a good idea (even prefixing with integers) as it can clobber user specified order (sans some special casing in the ordering).

campos-ddc:
Makes sense.

I used numbers to make sure that they will be executed in the order they are described in `environment.yml`, as far as I know, `find` in the `activate` script will return filenames sorted alphabetically.

How are you sorting scripts in your pull request? Alphabetically? Modification date?

tacaswell:
alphabetically.  I think they are currently coming back in random order,
but I am no longer sure why I think that (I don't remember if that PR was
due to something actually going wrong or just general paranoia that the
system can return the list in any order).

campos-ddc:
Alphabetically would work perfectly with my changes, there shouldn't be any conflict.

I commented in #66 , it looks like Windows is already sorting alphabetically (but it only lists files at the current dir, no recursion as we have in unix!).

Also, your PR is still valid, `find` in fact does not return files in any order.

tswicegood:
This one gives me pause.  I like the idea, but something we have to consider here is the complexity that it adds.  To this point, an environment spec is nothing more than one file.  If we add this, it's 1+n file(s).

To add this in its current state, we would need add error handling to `main_create` and `main_update` to ensure that an environment wasn't created unless it could be completely instantiated.  I'm not sure if that's a good or bad idea.

Another option that keeps this as a single file is to allow inlining of these activation scripts.  Something like this:

```yaml
- activate:
  - posix: |
      #!/bin/bash
      echo "I'm in a *nix system"
  - win: |
      echo "This is a bat file in Windows"
```

Thoughts?

campos-ddc:
IMO, a few points to consider:

* The complexity only exists if the user wants to have activate/deactivate scripts.
* Using activate/deactivate files (instead of text) fits perfectly with the way that pre-built packages deploy their activation scripts.
* `main_create` and `main_update` already suffer from this problem (they do not check that activation scripts installed by packages can be completely instantiated). If this is a behavior we want, it should be implemented regardless of adding activation scripts to `environment.yml`.
* Real script files can be executed independently, without going through conda. This can be useful for testing them.

I like the idea of having everything in a single file, but writing bash/bat scripts as text in a `.yml` file seems a bit odd. I guess it would work just fine for simple examples, but if I had a 200 line script, I wouldn't want it in text within another file (assuming we'll ever have a 200 line script being needed).

tswicegood:
I'm not sold on the need of one file, I just realize that it is a change we need to think about.  For example, @malev just added the ability to upload and download an environment to Anaconda Server.  With this change, now we have to adjust that to make sure that correct files are being sent to the correct places.

To address a couple of points:

> The complexity only exists if the user wants to have activate/deactivate scripts.

This is true to the user, but not us as maintainers of the package.  We have to be able to manage the complexity on our end as well and while it seems rather straight forward, there are moving parts (like dealing with "remote" packages that have been uploaded) that makes moving from a single file to a single file plus N other files unappealing.

> `main_create` and `main_update` already suffer from this problem (they do not check that activation scripts installed by packages can be completely instantiated). If this is a behavior we want, it should be implemented regardless of adding activation scripts to `environment.yml`.

These are not quite the same.  An `environment.yml` installs packages that may or may not actually do what they're supposed to.  That's left as an exercise to the package maintainer and pushes the complexity of verification further down the stack in order to keep conda-env simple.

That's not to say that there's a place for this.  I think there is, I think we just need to figure out the best way to tackle the problem.

---

To bring the conversation back to the main goal: the script you used as an example I assume is setting environment variables.  Is that correct?  If so, what about adding a section that specifies the variables.  Something like:

```yaml
- env:
  - ORACLE_HOME: "{CONDA_ENV_PATH}/opt/oracle"
```

Then instead of copying files over this would create a `_conda-environment-variables.sh` file or some such that sets all of these as environment variables.  Thoughts?  Is there a broader case where the file makes more sense?

campos-ddc:
Yes, I'm implementing another pull request right now that does just that:

```yaml
name: app

dependencies:
  - cogapp==2.4
  - colorama==0.3.3

environment:
  PYTHONPATH:
    - {{ root }}/source/python
```

For me, it will make the `activate` and `deactivate` scripts unnecessary.

That being said, having full control of a bash/bat script is more flexible than just setting environment variables, so I thought that the scripts could still be useful.

I had not considered recent changes in `conda env upload/download`.

Would you like to wait until I finish and open the other pull request? We can continue this discussion/abort this pull request if the other one is better.

tswicegood:
That sounds like a solid plan.  Thanks for tackling this @campos-ddc.

campos-ddc:
Just opened #88 , take a look there.

tacaswell:
I think there is a case for allowing full scripts.  I have a compelling use-case where I would like to have an environment starts several server processes to build a test stand environment (but just setting the ENVs is also great!).

campos-ddc:
I had a look at `conda env upload`, and could quite figure out why it necessary...

I thought that `environemnt.yml ` should always be a part of your repository, along with thew source code. Why would it be uploaded as a standalone file in binstar?

@malev , what use case did you have when implementing this?

malev:
@campos-ddc not every conda user has a repository. Many conda users are focused on jupyter notebooks. The idea behind `conda env upload` is to share environments between those users.

tadeu:
@malev These changes would be very useful, is there something else you need in order to merge, besides resolving the conflicts?

vatsan:
Is this going to be merged?

PR #95, opened 2015-06-08
Update Python 2-style exception handling to 'as'

jni:
This fixes a syntax error in Python 3. I don't actually understand why I can't trigger it normally, but if I cd into the `specs` directory and import it:

```python
>>> import yaml_file as yf
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/Users/nuneziglesiasj/anaconda/envs/py3/lib/python3.4/site-packages/conda_env/specs/yaml_file.py", line 16
    except EnvironmentFileNotFound, e:
```

### DIFF
--- a/conda_env/specs/yaml_file.py
+++ b/conda_env/specs/yaml_file.py
@@ -13,7 +13,7 @@ class YamlFileSpec(object):
         try:
             self._environment = env.from_file(self.filename)
             return True
-        except EnvironmentFileNotFound, e:
+        except EnvironmentFileNotFound as e:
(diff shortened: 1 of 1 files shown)


tswicegood:
:+1: -- not sure why that wasn't triggered already. :-/

ccordoba12:
It was also mentioned in the public Anaconda mailing list that just doing

    conda info

also shows this error.

asmeurer:
This is breaking things. We need to push it out soon.

Issue #98, opened 2015-06-15
`conda env upload` scary error


malev:
when uploading an environment already uploaded conda-env displays an ugly error message:
```
An unexpected error has occurred, please consider sending the
following traceback to the conda GitHub issue tracker at:

    https://github.com/conda/conda/issues

Include the output of the command 'conda info' in your report.


Traceback (most recent call last):
  File "/Users/mvanetta/miniconda/bin/conda-env", line 6, in <module>
    exec(compile(open(__file__).read(), __file__, 'exec'))
  File "/Users/mvanetta/code/conda/conda-env/bin/conda-env", line 5, in <module>
    sys.exit(main())
  File "/Users/mvanetta/code/conda/conda-env/conda_env/cli/main.py", line 65, in main
    return args_func(args, parser)
  File "/Users/mvanetta/miniconda/lib/python2.7/site-packages/conda/cli/main.py", line 208, in args_func
    args.func(args, p)
  File "/Users/mvanetta/code/conda/conda-env/conda_env/cli/main_upload.py", line 108, in execute
    info = uploader.upload(args.force)
  File "/Users/mvanetta/code/conda/conda-env/conda_env/utils/uploader.py", line 86, in upload
    raise exceptions.AlreadyExist()
conda_env.exceptions.AlreadyExist: The environment path already exists
```

malev:
Should be fixed in `2.2.2` release

PR #99, opened 2015-06-15
Display better error when env already uploaded

malev:
Problem: #98 
# Solution:

When trying to upload an environment that has already been uploaded show:
```
(conda-3)➜  tmp  conda env upload conda.io
Uploading environment conda.io to anaconda-server (https://api.anaconda.org)...
Error: The environment path already exists
```


tswicegood:
:+1:

PR #102, opened 2015-06-16
Add --force option to conda env create

campos-ddc:

### DIFF
--- a/conda_env/cli/main_create.py
+++ b/conda_env/cli/main_create.py
@@ -1,12 +1,14 @@
 from __future__ import print_function
 from argparse import RawDescriptionHelpFormatter
 import os
-import textwrap
 import sys
+import textwrap
 
 from conda.cli import common
 from conda.cli import install as cli_install
+from conda.install import rm_rf
 from conda.misc import touch_nonadmin
+from conda.plan import is_root_prefix
 
 from ..installers.base import get_installer, InvalidInstaller
 from .. import exceptions
@@ -58,6 +60,12 @@ def configure_parser(sub_parsers):
         default=None,
         nargs='?'
     )
+    p.add_argument(
+        '--force',
+        help='force creation of environment (removing a previously existing environment of the same name.',
+        action='store_true',
+        default=False,
+    )
     common.add_parser_json(p)
     p.set_defaults(func=execute)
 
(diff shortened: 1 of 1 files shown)


conda_env/cli/main_create.py
```diff
@@ -81,6 +88,8 @@ def execute(args, parser):
         common.error_and_exit(str(e), json=args.json)
 
     prefix = common.get_prefix(args, search=False)
+    if args.force and prefix != root_env_name and os.path.exists(prefix):
+        shutil.rmtree(prefix)
```
malev: Why don't you use `rm_rf` as in `conda env remove` https://github.com/conda/conda-env/blob/develop/conda_env/cli/main_remove.py#L105 ? Think this would be a more consistent approach

campos-ddc: Didn't know it existed.

I'll update the PR

asmeurer: Yes, please use `rm_rf`. `rmtree` is unreliable on Windows and `rm_rf` fixes many issues with it.

campos-ddc: Done.

malev: You can also benefit from `is_root_prefix` as in https://github.com/conda/conda-env/blob/develop/conda_env/cli/main_remove.py#L46

campos-ddc: Added `is_root_prefix` too.

conda_env/cli/main_create.py
```diff
@@ -81,6 +88,8 @@ def execute(args, parser):
         common.error_and_exit(str(e), json=args.json)
 
     prefix = common.get_prefix(args, search=False)
+    if args.force and not is_root_prefix and os.path.exists(prefix):
```
malev: `is_root_prefix` is a function that expects a prefix as an argument. Think you are missing that here.
Take a look at https://github.com/conda/conda/blob/master/conda/plan.py#L312
See usage in https://github.com/conda/conda-env/blob/develop/conda_env/cli/main_remove.py#L46

campos-ddc: derp.. fixing it.

malev:
@campos-ddc I can't merge this anymore. Would you mind updating the code from `upstream` to we can finally merge?

campos-ddc:
@malev Done.

Issue #97, opened 2015-06-10
`conda env export` does not include channels information


branning:
It seems that if you have installed conda packages from anything but the main channel, then you cannot create an environment from the output of `conda env export` due to the fact that the channel information is not included in the YAML output.

I'm using conda-env 2.2.1 with conda 3.13.0 with python 2.7.9 32-bit on Windows 7 64-bit. Here's a minimal example:

```
set MYENV=channeltest
conda create -n %MYENV% -c branning scandir -y
call activate %MYENV%
conda env export -f environment.yml
call deactivate
conda env remove -n %MYENV% -y
conda env create
```

If you run that, and you can install the Windows 32-bit scandir package from binstar, you'll see this error at the end:

    Error: No packages found in current win-32 channels matching: scandir 1.1 py27_0

But if you [edit environment.yml by adding the channel information](https://gist.github.com/branning/d75fa15e92f9ca7dc08c),

```
python fix_yml.py > environment_fix.yml
conda env create --file environment_fix.yml
```

it works as expected, or at least this is how I believe it should work.

malev:
There is more conversation about this in #103
Also we are considering expanding the way `export` works to include from what channel was `X` package installed. Still a thinking about this

msarahan:
closed by #103

PR #103, opened 2015-06-17
Feature/export channels

malev:
### Why?
When exporting channels information is missing:
`conda env export -n root`
```yaml
name: root
dependencies:
- binstar=0.11.0=py27_0
- certifi=14.05.14=py27_0
```
The environment.yaml file can't be used because it lack on the source of the packages (channels).

### Solution
Include channels (listed on `.condarc`) custom channels (`-c` flag):
`conda env export -c binstar --override-channels -n root`
```yaml
name: root
channels:
- binstar
dependencies:
- binstar=0.11.0=py27_0
- certifi=14.05.14=py27_0
```

There is an open bug talking about this as well #97


tswicegood:
You might want to talk with @chdoig about this a bit as she's one of the one's who's asked for this.  I believe the desired behavior here isn't explicitly setting channels at export time so much as gather the channels from Conda's metadata when exporting so you don't have to remember that channel `foobar` was used.

FWIW, the data you need is in the `{{ path to env }}/conda-meta/{{ package }}-{{ version }}-{{ build }}.json` file under `channel` if it exists.  To my knowledge there is no API for getting that data and no official documentation of what goes into that file.

malev:
* Didn't know about `conda-meta` it's great!
* I agree with you but I also believe this is a step towards reproducibility of environments.
* I don't want to make a broader change into `conda-env`. Not sure if we can include all that information into the `environment.yaml` file and how is the `conda env create` will respond to that. I saw this as an easy temporally fix
* In the near future I'd like to have namespaces into the `environment.yaml` file, something like:
```yaml
dependencies:
- javascript/nodejs
- javascript/bower
- binstar=0.11.0
```
And `conda-env create` will go to that particular channel to look for the package.

tswicegood:
+1 for the channel-specific package specification.  I know @asmeurer has given it some thought, but I don't know if anything's every been done on it.

msarahan:
Ping @malev can you revisit this and reassure me that it's ready?  I'll merge if you say it's good.

rain1024:
This is really helpful feature. Please merge it soon.

Issue #42, opened 2015-01-08
Command to get an environment.yml to share with someone 


chdoig:
Currently, the environment.yml file allows you to not specify a concrete version/build of the package, so if there are updates in that package and I share my environment.yml file with someone, they might get a different environment (different versions of the package), that might not be compatible. Although I could re-write the environment.yml to precisely specify the correct versions, it would be nice to have a command that generates a .yml file for you with what is currently downloaded in that environment (specific versions or build numbers). So you could share that env.yml and get the same conda environment that you are using.

tswicegood:
@chdoig does `conda env export --name=some-env` do what you need?  By default, it outputs a dependency list that's pinned to the explicit version and build string.  Issue #15 calls for adding a non-strict mode where you can export just versions or package names instead of the full spec.

chdoig:
I found the issue why export wasn't working for me: I had some packages from a binstar user or organization and `conda env export` doesn't add channels in the env.yml. I'll close this issues and open a more appropriate one. Thanks!

chdoig:
So, I'm reopening this issue because `conda env export` as is it is right now isn't fully letting me share my environment across operating systems (e.g OSX -> Linux64), for example in the Topic_Space project (https://github.com/ContinuumIO/topic_space/issues/7).

So, all the full list of what's in my environment might be platform specific, but what I've manually wanted to install to that environment might not be.

What are your thoughts on this @tswicegood? 
What do you think about having a way to export the environment that has just what you've explicitly installed in the environment? It probably doesn't solve all issues (if a package doesn't have a package for an OS), but, could it get us closer?

tswicegood:
I like the idea of only dealing with things that are explicitly installed.  Let's talk about this next week when I'm in the office and try to figure out a solution.

asmeurer:
A reminder that conda does keep track of this in the history file.

malev:
@chdoig can you tell me what were you thinking about this? tnx

chdoig:
@malev, I like the idea of being able to export just what you have explicitly installed in the environment. It probably should not be the default, but passing an option in the ```conda env export command```. I don't have ideas for the name of the option.

malev:
@chdoig, for now I have no idea how to do that and I think it's gonna be easier/better to have people creating their own `environment.yml` files on their own.

hovren:
I just wanted to echo that I too would find it convenient with some option to create a "shareable" environment with conda.

[...]

PR #104, opened 2015-06-17
export build numbers is optional

malev:
### Why?
If a package has been built for `win` and `linux` but for some reason, (more common than we think) the builds have different numbers, then the `environment.yaml` file won't be cross platform. On the other hand, we don't always need a specific build for a package, we only want the version number.

### Current situation
In order to have an `environment.yaml` file working in different platforms sometimes we have to edit the file and remove the build information from a specific package.

### Solution
Make exporting build information optional:
```sh
$ conda env export -n root
name: root
dependencies:
- certifi=14.05.14=py27_0
- clyent=0.3.4=py27_0
```
and
```sh
$ conda env export -n root --no-builds
name: root
dependencies:
- certifi=14.05.14
- clyent=0.3.4
```
The difference is the `no-builds` flag. cc @tswicegood @chdoig @tonyfast

### DIFF
--- a/conda_env/cli/main_export.py
+++ b/conda_env/cli/main_export.py
@@ -42,6 +42,14 @@ def configure_parser(sub_parsers):
         required=False
     )
 
+    p.add_argument(
+        '--no-builds',
+        default=False,
+        action='store_true',
+        required=False,
+        help='Remove build specification from dependencies'
+    )
+
     p.set_defaults(func=execute)
 
 
@@ -64,8 +72,7 @@ def execute(args, parser):
     else:
         name = args.name
     prefix = common.get_prefix(args)
-
-    env = from_environment(name, prefix)
+    env = from_environment(name, prefix, no_builds=args.no_builds)
 
     if args.file is None:
         print(env.to_yaml())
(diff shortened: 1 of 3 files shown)


groutr:
Why are build numbers included in yaml files?  Shouldn't we always use the latest build number for a package on it platform?

malev:
@groutr `certifi=14.05.14=py27_0` is how we currently generate `yaml` files. The build number is the problem and that's what I'm trying to remove with this PR.
To be honest, I believe it would be better if we remove the build number by default unless you explicitly ask for it.

groutr:
I was implying with my question that build numbers should be irrelevant.  Incrementing a build number should never change functionality of a package, except in the case of a broken package.  In every case I can think of, I can't come up with a scenario in which we would want to use an older build number for package.

chdoig:
+1 This will solve a lot of my issues with sharing environment.yaml files in different platforms.

I would also suggest having the one without the build number be the default export, rather than the other way around, but I'm happy with getting the functionality.

@groutr the build number is not completely irrelevant when you need exact reproducibility, explicitly know where each for package came from and identify the package with its URI, which includes the build number. If it was completely irrelevant, then they would be exactly the same and it shouldn't to define a conda package. I wouldn't remove it. I just think the functionality without the build number is going to be the general use case rather than with it, which is just going to be needed in very few occasions.

chdoig:
This PR is related with #42.

malev:
I'd go it two steps. First I'd release this and in a major release I'd change the current behavior to remove the build numbers by default

Issue #105, opened 2015-06-17
`conda env create --name` is broken


jim22k:
I want to create an environment from an existing environment.yaml file, but I want a different name for the environment than is listed in the file.

`conda env create --name custom_name` is broken and gives the following message:

    --name is deprecated. Use the following command instead:
        conda env create custom_name

Typing the recommended command doesn't work. It just does `conda env create` and completely ignores the `custom_name` argument.  I'm using conda 3.14.0.

I tried to use `conda create -n custom_name --file environment.yaml`, but that doesn't work either (see conda/conda#1360). How do I solve this problem?

joelhullcio:
@jim22k +1
The error is introduced by conda-env 2.2.0.   Important to note that the removal of this feature BROKE CLIENT SCRIPTS AT JPM.  We need to have a strong commitment to backward compatibility.  In addition, deprecation means a feature is going to be removed in the future but it should still work.  However, why would this feature be deprecated?  This is removing valuable functionality.  Don't we want to add features to conda instead of removing them?  Please add this feature back ASAP.

teoliphant:
@tswicegood Can you shed light on why this interface changed?

malev:
we change it to support creating environments from anaconda.org. Working on a fix right now

tswicegood:
@teoliphant the interface was changed to decrease the friction for creating a remote environment.  Writing out `conda env create --name tswicegood/foobar` seems off.  The intent here was to deprecate `--name`, not remove its functionality.  That was a regression as it was supposed to only display the warning that it was no longer the proper way, but would still work.

malev:
The fix has been merged, waiting the anaconda team to proceed with the release of `2.2.3`. I'm hoping tmr it will be done.

malev:
@jim22k could you update your `conda-env` and try again?

jim22k:
Just tried with conda-env v2.2.3 and it works. Thanks for the fast turnaround.

PR #106, opened 2015-06-18
Undo -n deprecation

malev:
Working on #105

### DIFF
--- a/conda_env/cli/main_create.py
+++ b/conda_env/cli/main_create.py
@@ -19,6 +19,7 @@ Create an environment based on an environment file
 example = """
 examples:
     conda env create
+    conda env create -n name
     conda env create vader/deathstar
     conda env create -f=/path/to/environment.yml
 """
@@ -41,9 +42,9 @@ def configure_parser(sub_parsers):
     p.add_argument(
         '-n', '--name',
         action='store',
-        help='environment definition [Deprecated]',
+        help='environment definition',
         default=None,
-        dest='old_name'
+        dest='name'
     )
     p.add_argument(
         '-q', '--quiet',
@@ -51,8 +52,8 @@ def configure_parser(sub_parsers):
         default=False,
     )
     p.add_argument(
-        'name',
-        help='environment definition',
+        'remote_definition',
+        help='remote environment definition',
         action='store',
         default=None,
         nargs='?'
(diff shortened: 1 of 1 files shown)


tswicegood:
Just reviewed from my phone, but this still deprecates --name in favor of the name without the opts but removes the message saying that. 

The original idea behind deprecating --name is that it should still work but display the warning message so that at some point it could be completely removed in the future when BC can be broken.

malev:
I'm not sure how do we should fix this. I'm feeling like deprecating `-n` wasn't the greatest idea. Maybe we should keep it the way it was and add a new flag for the `remote_definitions` as in:
```sh
$ conda env create --help
usage: conda-env create [-h] [-f FILE] [-n NAME] [-q] [--json]
                        [remote_definition]

positional arguments:
  remote_definition     remote environment definition

optional arguments:
  -n NAME, --name NAME  environment definition

examples:
    conda env create
    conda env create -n name
    conda env create vader/deathstar
    conda env create -f=/path/to/environment.yml
```
What do you think?

tswicegood:
I'm still -1 on `-n` as its weird to type `conda foo -n bar -y`.  That said, I can see a case where you need them both so maybe it's ok to keep around.

malev:
Agree with the weirdness but that's the way conda works in general, it's gonna be weirder to have conda using `-n` and conda-env not using it.

tswicegood:
Yeah -- my eventual hope was to deprecate/remove it there too, but I doubt that'll happen now. :-(

malev:
I still believe that:
`conda env create malev/stats-env` is the way to go!
And now it feels like we can also do:
`conda env create malev/stats-env -n new-env-name` and that seems to be pretty useful for of our users.

joelhullcio:
@tswicegood -n is a well established parameter across anaconda.  IF it is deprecated it should be deprecated in unison across all of conda for consistency.  I see your point on "weirdness" but it may be a premature optimization at this point.  Please just return the functionality to the previous state without deprecation.  Add other parameters to handle any new or special remote features.  Thanks.

tswicegood:
LGGT :+1:

eriknw:
I am very confused by this thread.

It appears you want to specify a remote environment definition obtained, by default, from anaconda.org.  Moreover, this is done from the *positional* argument.  I would expect arguments that specify remote resources to be explicitly called out, such as `conda env create -r malev/stats-env`.  The exception to this would be if the conda sub-command were something like `conda remote` or `conda env clone`.

What would most users expect the positional argument to `conda env create` do: (1) specify an environment yaml file, (2) new conda environment name, (3) remote conda name to install locally, or (4) some combination of these three?

I expect it to specify a conda environment definition.  Perhaps the positional argument can specify a local environment yaml file *or* a remote `{channel}/{envname}`.  Note that this is different from the original (broken) deprecation behavior, which said one could specify the new environment name as the positional argument.

To be clear, I think sharing environments via anaconda.org is an awesome feature to support.

malev:
@eriknw,
1) I'm gonna guess that `-f` flag exists because `conda` has a `-f` flag and make sense to have reproduce that here as well
2) yes! I agree and I think it's cool but as @joelhullcio pointed out "-n is a well established parameter across anaconda"
3) maybe! My inspiration on this was `git clone`. The original discussion is in https://github.com/conda/conda-env/pull/73
4) Good luck with this one!! Not simple to implement

All that said, I'm gonna merge this now because there is people missing this feature. We can continue this debate in an issue.

eriknw:
I remain somewhat confused, and I think usage of `conda env create` is unnecessarily complicated by kludging remote functionality into it.  I don't understand why (4) would not be simple to implement.

Thanks for adding `--name` functionality back--and not deprecating it--to override the new environment name.

PR #110, opened 2015-06-23
disable user site packages in conda envs

minrk:
It's surprising that activating a conda env doesn't disable packages that were installed in the global env with `--user`. This matches the default behavior of virtualenv, of creating an actually isolated env.

It would be preferable if this could be accomplished without reliance on an environment variable, but the detections in site.py seem to be pretty hardcoded to virtualenvs.

I'm not sure what, if anything, should be done for the windows scripts.

### DIFF
--- a/bin/activate
+++ b/bin/activate
@@ -77,6 +77,12 @@ else
     return $?
 fi
 
+if [[ ! -z "${PYTHONUSERSITE+x}" ]]; then
+    export CONDA_PYTHONUSERSITE="$PYTHONUSERSITE"
+fi
+# disable user site-packages
+export PYTHONUSERSITE=0
+
 # Load any of the scripts found $PREFIX/etc/conda/activate.d
 run_scripts "activate"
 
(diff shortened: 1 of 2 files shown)


malev:
hello @minrk how can I test this?

minrk:
@malev 

Simplest way to test:

1. activate an env with `source activate <name>`
2. check output of: `python -m site`. It should have `ENABLE_USER_SITE: False`

Test the actual behavior that is to be avoided:

1. conda root env
2. install a package with `--user`, e.g. `pip install --user simplejson
3. activate an env with just Python, and try to `import simplejson`. It should fail.

minrk:
I discovered that there's a better way to do this that will work on Python 3, without relying on environment variables. I still think the env should be set for Python 2, but a `pyvenv.cfg` file can be created, which site.py will check.

Creating a `pyvenv.cfg` with the contents:

    include-system-site-packages=false

in the root of the env (e.g. `~/conda/envs/name/pyvenv.cfg`), user-site will be disabled. I don't know if `conda-env` is the right place for that addition, or the Python conda package, which is not open source. It seems like `conda-env` might be the place to do it, but I'm not sure.

PR #111, opened 2015-06-23
conda env attach

malev:
#### Problem
Notebooks need an environment to run. If we are going to share notebooks in different platforms, we need a way to share the environment with the notebook.

#### Solution
Attach the current environment into the notebook's metadata. `conda env attach` will grab the curren environment and store it into the notebook's metadata. It will also allow to store remote environments (hosted in anaconda.org) for instance.

```sh
$ conda env attach --help
usage: conda-env attach [-h] (-n NAME | -r REMOTE) [--force] [--json] notebook

Embeds information describing your conda environment
into the notebook metadata

positional arguments:
  notebook              notebook file

optional arguments:
  -h, --help            show this help message and exit
  -n NAME, --name NAME  local environment definition
  -r REMOTE, --remote REMOTE
                        remote environment definition
  --force               Replace existing environment definition
  --json                Report all output as json. Suitable for using conda
                        programmatically.

examples:
    conda env attach -n root notebook.ipynb
    conda env attach -r user/environment notebook.ipynb
```
cc @tonyfast, @chdoig

### DIFF
--- a/conda_env/cli/main.py
+++ b/conda_env/cli/main.py
@@ -28,6 +28,7 @@ environment, please open a bug report at:
     else:
         raise e
 
+from . import main_attach
 from . import main_create
 from . import main_export
 from . import main_list
@@ -48,6 +49,7 @@ def create_parser():
     p = argparse.ArgumentParser()
     sub_parsers = p.add_subparsers()
 
+    main_attach.configure_parser(sub_parsers)
     main_create.configure_parser(sub_parsers)
     main_export.configure_parser(sub_parsers)
     main_list.configure_parser(sub_parsers)
--- a/conda_env/cli/main_attach.py
+++ b/conda_env/cli/main_attach.py
@@ -56,9 +56,9 @@ def configure_parser(sub_parsers):
 def execute(args, parser):
     if args.name is not None:
         prefix = common.get_prefix(args)
-        content = from_environment(args.name, prefix).to_yaml()
+        content = from_environment(args.name, prefix).to_dict()
     else:
-        content = "remote: {}".format(args.remote)
+        content = {'remote': args.remote}
 
     print("Environment {} will be attach into {}".format(args.name, args.notebook))
     nb = Notebook(args.notebook)
--- a/conda_env/cli/main_attach.py
+++ b/conda_env/cli/main_attach.py
@@ -23,10 +23,17 @@ def configure_parser(sub_parsers):
         help=description,
         epilog=example,
     )
-    p.add_argument(
+    group = p.add_mutually_exclusive_group(required=True)
+    group.add_argument(
         '-n', '--name',
         action='store',
-        help='environment definition',
+        help='local environment definition',
+        default=None
+    )
+    group.add_argument(
+        '-r', '--remote',
+        action='store',
+        help='remote environment definition',
         default=None
     )
     p.add_argument(
@@ -46,13 +53,11 @@ def configure_parser(sub_parsers):
 
 
 def execute(args, parser):
-    print(args)
-    if args.name is None:
-        args.name = current_env()
+    if args.name is not None:
         prefix = common.get_prefix(args)
         content = from_environment(args.name, prefix).to_yaml()
     else:
-        content = "name: {}".format(args.name)
+        content = "remote: {}".format(args.remote)
 
     print("Environment {} will be attach into {}".format(args.name, args.notebook))
     nb = Notebook(args.notebook)
--- a/conda_env/cli/main_attach.py
+++ b/conda_env/cli/main_attach.py
@@ -42,7 +42,6 @@ def configure_parser(sub_parsers):
         default=None
     )
     common.add_parser_json(p)
-
     p.set_defaults(func=execute)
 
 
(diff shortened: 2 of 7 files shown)


ahmadia:
Excellent!  I haven't had a chance to look at this closely, but from an API perspective, it might be good not to assume that the attached object is an IPython notebook.  Perhaps in the documentation give it as an example of a supported type, and raise a type error if the file is not an .ipynb file or not a valid notebook

asmeurer:
It's not clear to me what "attach" means, and the command help is not illustrative.

ahmadia:
In this case, it means "embed information describing the environment into the notebook metadata".  @teoliphant was +1 on the verb, but perhaps there's a better way to describe it?

damianavila:
I did not go deep in the implementation, but I don't see any use of the IPython API to deal with the notebook metadata content... is this a design choice?

malev:
@asmeurer thanks! I'm gonna work on the description
@damianavila I don't want to add `ipython-notebook` as a dependency, that's why I'm treating the notebook as a `json` document. Let me know what do you think about that.
Waiting on comments from @chdoig and @tonyfast

malev:
@asmeurer I took @ahmadia suggestion to improve the description. I'd love your input

malev:
Also I'm starting to think that we will need a `conda env detach` command. OR A BETTER NAME!

damianavila:
>@damianavila I don't want to add ipython-notebook as a dependency, that's why I'm treating the notebook as a json document. Let me know what do you think about that.

I am a little worried that when things go complex, you have to rebuild all the machinery...
Since IPython is splitting in multiple repos... and one repo was specifically splitted to deal with notebook formats (https://github.com/jupyter/nbformat), I think it could be valuable to use ipython for now and when the split is complete (in a few weeks), just migrate to use nbformat which would be a little dependency...

ahmadia:
I'm -1 on adding IPython or nbformat as a dependency since conda itself should be as lightweight as possible, and the IPython notebook structure changes so infrequently.

damianavila:
>I'm -1 on adding IPython or nbformat as a dependency since conda itself should be as lightweight as possible

Maybe an optional dependency? The lightweight nature of conda is a good point, I agree...

>IPython notebook structure changes so infrequently.

Well, not so infrequently, and we had a new format relatively close in time...

For example... the current `tests/support/notebook.ipynb` in this PR in not nb.v4 compliant... it is detected as an old format v2... Then you have the `tests/support/notebook-with-env.ipynb`, which is also detected as v2, but when it is transformed to v4, the metadata is lost (maybe a bug, we should investigate), but if you create the v4 format and add the environment field, it is not lost at all... this example illustrate why using the IPython/nbformat machinery prevent you to hit some problems... in any case, just a suggestion... I understand that making IPython is maybe too much as a conda dependency (even optional) but I would think about using nbformat in the future... it is light (a few dependencies)...

malev:
Since `conda-env attach` works with notebooks, having `ipython notebook` as a dependency does not sounds out of the place. I'm thinking about having it as an optional dependency. Said that, I believe that using the [nbformat module](http://ipython.org/ipython-doc/3/api/generated/IPython.nbformat.html#module-IPython.nbformat) is the best alternative.

groutr:
would embed/unembed be better word choices

malev:
@groutr @teoliphant liked `attach` that's why we went with this one. There is an email thread with more background I'll forward it to you

groutr:
Oh.  The tone I got from @ahmadia last comment on the subject felt like it was still open for discussion.

ahmadia:
@groutr - I don't think either Travis or I are, heh, "attached" to it.  I don't like add/remove (believe they're already being used), and embed/unembed aren't as great antonyms.  I think we should just proceed as is.  Thanks for the suggestion though :)

malev:
think about storing JSON instead of YAML

ahmadia:
@malev - agreed.  The YAML environment should compile to JSON.  I think this was true in the PR/branch I made?

tonyfast:
#### IPython dependency

Adding the conda environments to the ``nbformat`` provides a lot of value to the notebook over time.  In the ``nbformat``, IPython is _not_ a dependency because this added conda feature simply augments metadata.  Some users may never open the notebook in the kernel, but they might want to know what environment is used simply from viewing the notebook.

#### The API

Chances are this feature will be most valuable as an add-on to the notebook.  So the phrasing might not be all that important in the UI.

This is a new feature that many folks are far from adopting which means some folks might need to play around with the language first.  Some suggestions I may make though:

* ``conda env nbformat file.ipynb`` - append the environment to the ``nbformat``
* ``conda env share file.ipynb`` - __share__ the environment in the notebook
* ``conda env export file.ipynb`` - __export__ the environment to the notebook.

malev:
**current status**
notebooks are being stored with `nbformat`, that allows us to work with different `nbnotebook`'s versions and store them in a standardised way. On the other hand we have a new dependency, but it's not a mandatory dependency. You can use `conda-env` without having `ipython` unless you need to work with notebooks. I believe this is the right approach.

tonyfast:
You can even copy and paste in the notebook.  Edit > Edit Notebook Metadata

ahmadia:
> current status notebooks are being stored with nbformat, that allows us to work with different nbnotebook's versions and store them in a standardised way. On the other hand we have a new dependency, but it's not a mandatory dependency. You can use conda-env without having ipython unless you need to work with notebooks. I believe this is the right approach.

Can we try to import nbformat then drop back to a raw JSON approach?

ahmadia:
Or at least report back `conda env attach requires that you install nbformat` with the following command, before you use it?

tonyfast:
You don't need ``nbformat `` to do any of this.  The notebook is in the ``nbformat`` already and this feature simply adds a patch to the json.

ahmadia:
> You don't need nbformat to do any of this. The notebook is in the nbformat already and this feature simply adds a patch to the json.

See the above discussion, you need nbformat because the JSON schema changes between notebook versions and nbformat (tries to) handle these version changes for you.

tonyfast:
The ``nbformat`` version is in the [notebook Top-Level structure](https://ipython.org/ipython-doc/3/notebook/nbformat.html#top-level-structure).  Is that not enough?

PR #116, opened 2015-07-08
Add %PREFIX%\Library\bin to path when activating in Windows

### LINKED ISSUE #115: %LIBRARY_BIN% not added to %PATH% when activating an environment in Windows
From what I can see, activating an environment in Windows will add the following directories to `%PATH%`:

* `%PREFIX%`
* `%PREFIX%/Script` (or `%SCRIPTS%` during the build process)

But `%PREFIX%/Library/bin` (or `%LIBRARY_BIN%` during the build process) is *not* added.

Is this intentional?

campos-ddc:
Fixes #115


malev:
ok, this needs to be updated and also, when deactivating we need to remove it from the path. Can you do that @campos-ddc please?

PR #118, opened 2015-07-08
Fix problem with PROMPT not resetting when activating within another env

### LINKED ISSUE #117: activate.bat while in another environment mangles %PROMPT%
Calling `activate.bat` while in another environment will correctly remove the old environment from `%PATH%`, but it does not fix `%PROMPT%`

This leads to cases like this:

```
campos@home C:\Users\campos

$ activate myenv
Activating environment "myenv"...
```

```
[myenv] campos@home C:\Users\campos

$ activate myenv
Deactivating environment "myenv"...
Activating environment "myenv"...
```

```
[myenv][myenv] campos@home C:\Users\campos
:: Double env prompt!
```

campos-ddc:
Fixes #117


gdementen:
The fix I proposed in #69 is IMO better (see my comment there).

PR #119, opened 2015-07-09
conda env export  has no-builds as default

malev:
* `--no-builds` deprecated
* `--with-bilds` new flag
* `conda env attach` adds environment without builds name


groutr:
I think there is a better solution than stripping build numbers.  It would be a whole more useful if conda env matched to varying levels of strictness when replicating an environment.  The user could control whether it matches just the package name, package name + version, or package name + version + build number.

This would mean that environment.yml should include build numbers.  I think it is better to adjust the interpretation of the these values rather than strip them out entirely.

Thoughts?

malev:
**on hold** for now

PR #121, opened 2015-07-10
Feature/env from notebooks

malev:
# Problem
You can attach an environment into a notebook but you can't do anything with it.

# Solution
Create environment from notebooks with an environment definition stored on it.

# Use
`conda env create notebook.ipynb`

### DIFF
--- a/conda_env/specs/notebook.py
+++ b/conda_env/specs/notebook.py
@@ -1,0 +1,29 @@
+try:
+    from IPython import nbformat
+except ImportError:
+    nbformat = None
+from ..env import Environment
+
+
+class NotebookSpec(object):
+    msg = None
+
+    def __init__(self, name=None, **kwargs):
+        self.name = name
+
+    def can_handle(self):
+        try:
+            self.nb = nbformat.reader.reads(open(self.name).read())
+            return 'environment' in self.nb['metadata']
+        except AttributeError:
+            self.msg = "Please install IPython notebook"
+        except IOError:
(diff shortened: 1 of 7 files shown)


conda_env/specs/notebook.py
```diff
@@ -0,0 +1,35 @@
...
+            self.nb = nbformat.reader.reads(open(self.name).read())
+            return 'environment' in self.nb['metadata']
+        except AttributeError:
+            self.msg = "Please install IPython notebook"
+        except IOError:
+            self.msg = "{} does not exist o can't be accessed".format(self.name)
+        except (nbformat.reader.NotJSONError, KeyError) as e:
+            print(e)
+            self.msg = "{} does not looks like a notebook file".format(self.name)
+        return False
```
groutr: I would expect a function that returns False would also be capable of returning True.  Maybe return None if unsuccessful?  In that case I would rename the function get_env() or something similar which returns the environment if available, otherwise None.

groutr:
is metadata a tag that we introduce to the notebook, or is it already define in the notebook specification?  Would we want to use something that is more specific to conda-env (like metadata/conda/environment); that way it would be easier to safely strip conda metadata?

conda_env/specs/notebook.py
malev: It does return `True` or `False`. Lines 19 and 27

malev:
@groutr `metadata` is specific to notebooks.
The other change will require a change in the way we specify an environment in a notebook.

PR #125, opened 2015-07-14
Added command 'conda env print' (receives environment.yml file)

campos-ddc:
Prints content from `environment.yml` files, this also handles `includes`, `jinja2` parsing and any other shenanigans that might be invented for `environment.yml` files.


malev:
What's the difference between this and a simple `cat`?

campos-ddc:
In the current version of conda-env, not much (it just adds --json support)

I made this to go along with #88, where not all environment information is in a single environment.yml file.

PR #126, opened 2015-07-14
Made it possible for --name to override name in environment.yml

campos-ddc:


campos-ddc:
Something went wrong after testing this on current HEAD

PR #128, opened 2015-07-15
Fix CONDA_ENV_PATH in windows

campos-ddc:
It was always setting it to ANACONDA_ENVS instead of the just-activated
environment


malev:
Don't understand what is the advantage of moving `set CONDA_ENV_PATH=%ANACONDA_ENVS%\%CONDA_DEFAULT_ENV%` down

campos-ddc:
It's now after `CONDA_DEFAULT_ENV` has been set with the name value.

campos-ddc:
New value* (sorry, can't edit comments on phone)

PR #148, opened 2015-07-31
Fix prompt issue in windows

malev:
When activating multiple environments win windows and then deactivating, prompt goes crazy. Taken from #69 Thanks to @gdementen


gdementen:
Haven't tested this, but I believe this is still buggy. The changes from #118 should probably be reverted, because otherwise, if you activate with another env already active, it first clears CONDA_OLD_PROMPT then set it again, which effectively has the same problem than #118.

PR #152, opened 2015-08-11
Add Library\bin to path in windows

malev:
Should close #116


campos-ddc:
@malev , I've had some issues with `Library/lib` not being in the `%PATH%` (some packages put `.dll` files there).

I think that it would be better for conda to include `Library/lib` too, or else these linux-style packages would have to move their dlls to /bin.

What do you think?

malev:
@campos-ddc, yeah, I guess. Just create a new issue or submit a PR

PR #160, opened 2015-08-15
Using Jupyter 4.0

malev:


ccordoba12:
@malev, I couldn't help to notice that you changed `ipython` for `jupyter` in this PR. But if conda-env depends only on the notebook part of Jupyter, I think you should just use the new `notebook` package instead of `jupyter`.

The thing is `jupyter` is a metapackage which installs not only the notebook but also the `qtconsole` and `console` frontends, which it seems you really don't need in this case :-)

malev:
@ccordoba12 `jupyter` is an optional dependency from conda-env, you don't need it at first glance. You only need `jupyter` when you are dealing with notebooks. If that's the case it's very possible that you will have jupyter already installed.

ccordoba12:
Ok, I didn't know that. But my suggestion was for `conda-env` to (optionally) depend *directly* on the `notebook` package instead of on `jupyter`.

I mean, the idea of Jupyter is not only to rename IPython to reflect its language agnostic nature, but also to allow a more fine-grained control over its multiple components.

For example, Spyder will depend (from now on) on the `qtconsole` package because that's the only Jupyter frontend that we currently use. The `jupyter` package brings several other components (like the notebook) that we don't really need :-)

damianavila:
:+1: to @ccordoba12 commets...

PR #161, opened 2015-08-18
Start using nbformat instead of jupyter

malev:
Coming from comments in #160


malev:
@ccordoba12 would you mind taking a look here?

damianavila:
Should be OK if we want to provide the minimal dependency to work with the notebook format...

ccordoba12:
:+1: from me too :-)

PR #172, opened 2015-09-27
Add support for creating envs from requirements.txt files via pip

dan-blanchard:
I found myself frequently wishing this existed, so I added it.

It will use conda packages if available, and fallback to pip otherwise.

There was a bit of nonsense necessary to get editable packages working
properly, but it all works.

### DIFF
--- a/conda_env/cli/main_create.py
+++ b/conda_env/cli/main_create.py
@@ -80,6 +80,8 @@ def execute(args, parser):
         # FIXME conda code currently requires args to have a name or prefix
         if args.name is None:
             args.name = env.name
+        else:
+            env.name = args.name
 
     except exceptions.SpecNotFound as e:
         common.error_and_exit(str(e), json=args.json)
--- a/conda_env/env.py
+++ b/conda_env/env.py
@@ -5,6 +5,13 @@ from collections import OrderedDict
 from copy import copy
 from io import open
 
+# Try to import PipSession to support new pips
+try:
+    from pip.download import PipSession
+except ImportError:
+    pass
+from pip.req import parse_requirements
+
 # TODO This should never have to import from conda.cli
 from conda import install
 from conda.api import get_index
@@ -61,27 +68,30 @@ def from_yaml(yamlstr, **kwargs):
     return Environment(**data)
 
 
-def from_requirements_txt(reqs_file, **kwargs):
+def from_requirements_txt(filename, **kwargs):
     """Load and return an ``Environment`` from a given ``requirements.txt``"""
     pip_reqs = []
     dep_list = []
     r = Resolve(get_index())
-    for line in reqs_file:
-        line = line.strip()
+    try:
+        parsed_reqs = list(parse_requirements(filename))
+    # Newer versions of pip require session kwarg
+    except TypeError:
+        parsed_reqs = list(parse_requirements(filename, session=PipSession()))
+    while parsed_reqs:
+        req = parsed_reqs.pop()
+        req_str = str(req.req)
+        if req.editable:
+            pip_reqs.append('-e {}'.format(str(req.link)))
         # If it's not an editable package, check if it's available via conda
-        if not line.startswith('-e'):
-            line = line.split('#', 1)[0].strip()
-            if line:
-                try:
-                    # If package is available via conda, use that
-                    r.get_pkgs(MatchSpec(common.arg2spec(line)))
-                    dep_list.append(line)
-                except NoPackagesFound:
-                    # Otherwise, just use pip
-                    pip_reqs.append(line)
-        # Editable packages just use pip
-        elif line:
-            pip_reqs.append(line)
+        else:
+            try:
+                # If package is available via conda, use that
+                r.get_pkgs(MatchSpec(common.arg2spec(req_str)))
+                dep_list.append(req_str)
+            except NoPackagesFound:
+                # Otherwise, just use pip
+                pip_reqs.append(req_str)
     dep_list.append('pip')
     dep_list.append({'pip': pip_reqs})
     data = {'dependencies': dep_list}
@@ -95,8 +105,7 @@ def from_file(filename):
     if not os.path.exists(filename):
         raise exceptions.EnvironmentFileNotFound(filename)
     if filename.endswith('.txt'):
-        with open(filename, encoding='utf-8') as fp:
-            return from_requirements_txt(fp, filename=filename)
+        return from_requirements_txt(filename)
     else:
         with open(filename, 'rb') as fp:
             return from_yaml(fp.read(), filename=filename)
--- a/conda_env/env.py
+++ b/conda_env/env.py
@@ -1,12 +1,15 @@
 from __future__ import absolute_import, print_function
+
+import os
 from collections import OrderedDict
 from copy import copy
-import os
+from io import open
 
 # TODO This should never have to import from conda.cli
-from conda.cli import common
-from conda.cli import main_list
 from conda import install
+from conda.api import get_index
+from conda.cli import common, main_list
+from conda.resolve import NoPackagesFound, Resolve, MatchSpec
 
 from . import compat
 from . import exceptions
@@ -50,8 +53,38 @@ def from_environment(name, prefix, no_builds=False):
 
 
 def from_yaml(yamlstr, **kwargs):
-    """Load and return a ``Environment`` from a given ``yaml string``"""
+    """Load and return an ``Environment`` from a given ``yaml string``"""
     data = yaml.load(yamlstr)
+    if kwargs is not None:
+        for key, value in kwargs.items():
+            data[key] = value
+    return Environment(**data)
+
+
+def from_requirements_txt(reqs_file, **kwargs):
+    """Load and return an ``Environment`` from a given ``requirements.txt``"""
+    pip_reqs = []
+    dep_list = []
+    r = Resolve(get_index())
+    for line in reqs_file:
+        line = line.strip()
+        # If it's not an editable package, check if it's available via conda
+        if not line.startswith('-e'):
+            line = line.split('#', 1)[0].strip()
+            if line:
+                try:
+                    # If package is available via conda, use that
+                    r.get_pkgs(MatchSpec(common.arg2spec(line)))
+                    dep_list.append(line)
+                except NoPackagesFound:
+                    # Otherwise, just use pip
+                    pip_reqs.append(line)
+        # Editable packages just use pip
+        elif line:
+            pip_reqs.append(line)
+    dep_list.append('pip')
+    dep_list.append({'pip': pip_reqs})
+    data = {'dependencies': dep_list}
     if kwargs is not None:
         for key, value in kwargs.items():
             data[key] = value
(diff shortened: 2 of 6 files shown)


dan-blanchard:
I tweaked this a bit so that it now pulls in all dependencies of packages in requirements.txt and checks to see if those are available as conda packages.  That way people don't end up with pip-installed numpy and other stuff like that.

dan-blanchard:
All tests are now passing, so this is good for review.

dan-blanchard:
I have no idea why this build is failing on Travis when it works fine on my machine.

conda_env/env.py
```diff
@@ -58,11 +72,75 @@ def from_yaml(yamlstr, **kwargs):
...
 
 
+def get_all_pip_dependencies(requirements_path, temp_dir='_delete_when_done'):
+    """
+    Use pip to gather all of the packages that would be installed for the given
+    requirements.txt file.
+    """
+    pip_cmd = ('pip', 'install', '--src', temp_dir, '--download', temp_dir,
+               '--no-use-wheel', '-r', requirements_path)
+    try:
```
msarahan: See comment elsewhere.  Please use isdir to check if dir exists, then create if necessary rather than try/except.

conda_env/env.py
```diff
@@ -58,11 +72,75 @@ def from_yaml(yamlstr, **kwargs):
...
+            data[key] = value
+    return Environment(**data)
+
+
 def from_file(filename):
     if not os.path.exists(filename):
         raise exceptions.EnvironmentFileNotFound(filename)
-    with open(filename, 'rb') as fp:
-        return from_yaml(fp.read(), filename=filename)
+    if filename.endswith('.txt'):
```
msarahan: This seems a little too broad of a catch-all.  Since the standard name is requirements.txt, would it be OK to match just that?  Otherwise, I can see this leading to some interesting bug hunts.

conda_env/env.py
```diff
@@ -58,11 +72,75 @@ def from_yaml(yamlstr, **kwargs):
...
+        return from_requirements_txt(filename)
+    else:
```
msarahan: Here, rather than a catch-all else statement, replace with the standard for yaml input (I think this is environment.yaml?)

If we don't match standard names, then we need to fall out, or provide a way for people to manually specify what kind of file they are providing.  This means modifying this function to accept another argument to specify this.

msarahan:
Sorry I don't have time to help you troubleshoot Travis right now.  I would like to merge this if you can get it past Travis.  Please see my other comments.  This is a nice addition - thanks for your work on it.

conda_env/env.py
dan-blanchard: I feel like people frequently have things like `dev-requirements.txt` and `test-requirements.txt`, so maybe we could do:

```python
if filename.endswith('requirements.txt'):
```

conda_env/env.py
dan-blanchard: I feel like checking for specific file names is a little overly strict.  If people are passing in file names as arguments part of `conda env create -n foo -f FILENAME`, it seems strange to restrict what the filenames must be.  Why not just assume that `.yml` or `.yaml` files should be loaded as such, and `.txt` files should be treated as pip/setuptools requirements files?

msarahan: This is fine, until someone else defines a new, shiny format for their package management tool that also happens to use yaml.  Making it explicit outside of standard filenames is safer over the long haul.  I think your idea of "endswith" is a good compromise between flexibility and prevention of future unexpected failure.

dan-blanchard: The `--file` argument is making it explicit.  The user is explicitly saying, "This is the file that defines my environment."  The default value for that argument is `environment.yml`, but anything else that is assigned to it is obviously what the person wants to create their environment from.  If in the future someone comes up with a new package format that use YAML that they would like to be supported by `conda env`, they're going to need to create pull request to add it anyway, so I don't see how just checking based on the extension right now (and raising an error if the extension isn't in `{'txt', 'yml', 'yaml'}`) is dangerous in any way.

malev:
@dan-blanchard what's gonna happen if in the same directory you have an `environment.yml` file and a `requirements.txt` file?
Also, I think you sould take a look at https://github.com/conda/conda-env/tree/develop/conda_env/specs
You could have a new spec called `requirements_txt_spec` (maybe)
And this one should return a `env` like object.
This is how we architecture the different ways to create an environment.
make sense?

dan-blanchard:
> @dan-blanchard what's gonna happen if in the same directory you have an environment.yml file and a requirements.txt file?

The `environment.yml` file will be used by default (since it's the default value for the `--file` argument to `conda env create`), and `requirements.txt` will be used if you specify it as the file you want to use with `--file`.

> You could have a new spec called requirements_txt_spec (maybe)

I can certainly do that if that's what y'all want me to, but I went with this since it is so similar to what is happening for the pip requirements in `environment.yml`.  I wasn't sure if the extra spec was really necessary, but obviously I'm not intimately familiar with the conda-env code.

conda_env/env.py
msarahan: It's not dangerous, but the error messages that someone will encounter in the future will not be helpful.  All I am really asking is that you provide a helpful error message if loading fails.  Loading should never fail, so long as files follow known standards (environment.yml and requirements.txt).  YAML and txt as file extensions happen to currently capture the current range of possibilities.  When someone adds something new (and you know they will, and I hope they use YAML, because I think it is technically good), I'd like users to meet a friendly error message along the lines of:

```
The specified file is not recognized as a conda-env input.  Conda-env currently support conda-env's environment.yaml specification, and pip's requirements.txt specification.  Please file an issue for the new format you're trying to use at https://github.com/conda/conda-env/issues
```

Rather than something obscure like:
```KeyError: ...```

**Comment by malev** [comment]
@dan-blanchard I think going with specs is a better approach because you'll take advantage of the architecture already in place for creating new environments.

File: conda_env/cli/main_create.py [hunk 4]
```diff
@@ -80,6 +80,8 @@ def execute(args, parser):
         # FIXME conda code currently requires args to have a name or prefix
         if args.name is None:
             args.name = env.name
+        else:
```
malev: don't fully understand what's going on here

conda_env/env.py
```diff
@@ -50,7 +64,7 @@ def from_environment(name, prefix, no_builds=False):
 
 
 def from_yaml(yamlstr, **kwargs):
-    """Load and return a ``Environment`` from a given ``yaml string``"""
+    """Load and return an ``Environment`` from a given ``yaml string``"""
```
malev: nice catch!

malev:
I'm starting to feel like this might be outside the scope of the project. What does @msarahan thinks about?

msarahan:
I don't think it's out of scope, necessarily.  I see it as a good compatibility layer that may make conda a more compelling platform.  The best platform isn't conda or pip, it's both.  I don't fully understand the long-term support implications, though.  It sort of joins conda to pip, and requires you to keep supporting pip once people start using this feature.  What happens if/when pip is replaced down the line?  You'll have to support pip likely longer than pip itself is supported.  I don't think this issue is a deal breaker.  I just think we need to discuss long-term implications, and figure out if any engineering solution can reduce future risk.

malev:
@msarahan that's my concern.
On the other hand I like it ... 
I'm gonna try to put some time on run tests and try out this PR

conda_env/cli/main_create.py
dan-blanchard: Without this the environment won't get a name, because `requirements.txt` files cannot contain the name of the environment.

conda_env/installers/pip.py
```diff
@@ -1,10 +1,19 @@
...
 from conda.cli import main_list
 
 
 def install(prefix, specs, args, env):
-    pip_cmd = main_list.pip_args(prefix) + ['install', ] + specs
+    # This mess is necessary to get --editable package sent to subprocess
+    # properly.
+    specs = list(chain(*[s.split() if '-e' in s else [s] for s in specs]))
+    # Directory where pip will store VCS checkouts for editable packages
+    src_dir = join(config.envs_dirs[0], env.name, 'src')
```
malev: looks like `config.envx_dirs[0]` is the location of the root environment.
how is this gonna work when we want to install into some other environment? (like the one we have just created)

dan-blanchard: `config.envs_dirs[0]` should be `$CONDA_ROOT/envs`, and I'm joining that with the name of the environment and `src`, so the final path will be `$CONDA_ROOT/envs/$CONDA_ENV/src`.  That's exactly what we want.

githubfun:
creating envs from requirements.txt files is very important. Often I am in a project which uses virtualenv and has a requirements.txt file. To reproduce the environment it is hard to use conda because it doesn't import requirements.txt.

dan-blanchard:
@githubfun, I've been successfully using my branch since September without any issues, so if this is something you need as much as I do, I suggest you just clone my branch and install it in your conda root environment with `python setup.py develop`.

dan-blanchard:
Any movement on adding this officially? I was talking to somebody at work about how they can use conda with requirements.txt files now and then I remembered it's only with my fork.

cowlicks:
One thing I haven't seen mentioned here, `conda install` has a `--file` flag which can take a requirements.txt file. This is only helpful when everything in the requirements.txt is has a conda package. But it is still probably useful in some cases @githubfun @dan-blanchard . Sorry if y'all were already aware of this.

However using  `conda install --file` is usually a pretty bad experience since most of the time there is not a conda package for everything in the requirements.txt. Trying to use `conda install --file` makes me want this.

At the end of the day what I (and probably others) want is a consistent workflow across projects, whether they use a requirements.txt or environment.yml. Something like

1. `cd project-name`
2. `conda env create`
3. `source activate project-name`

Which is what this PR would let me 

So yeah +1 to this.

@malev how does #203 address this?

AlJohri:
@malev any word on how to get this merged in? what can we do to help?

dan-blanchard:
> This is only helpful when everything in the requirements.txt is has a conda package.

Yeah, the main advantage of the approach here is that this will support editable packages on GitHub and the like, which are currently not supported by `conda env`.

conda_env/env.py
lukeyeager: FWIW, I'd like to match `*requirements*.txt` so that files like `requirements_test.txt` will work too.

mlaprise:
hey @dan-blanchard, like you probably know, I'm really interested by this. Are you using this approach in production ? As of today, is there any other ways to achieve the same goal ?

dan-blanchard:
hey @mlaprise, we still aren't using conda in production at Parse.ly, but I use this at least once a week for development. I don't know of any other way to achieve the same goal.

jni:
FWIW I would very much like to see this addition merged. I have wanted it so many times.

nicoddemus:
@jni this repository has been deprecated, if there's still interested in this PR it should be re-opened in `conda/conda`.

@ccordoba12 @kalefranz perhaps using the GH API to post a comment in all open issues and PRs to let people now that the developement has been moved to `conda/conda` is a good approach to prevent confusion?

PR #182, opened 2015-10-14
Fix/activate behaviour on windows

wulmer:
This PR is a follow-up on my original pull request [#913](https://github.com/conda/conda/pull/913) in the `conda/conda` repository. As `conda` has been split up into `conda` and `conda-env`, parts of the original pull request must now go into this repo, while most of the work is done in a new pull request on the `conda/conda` repo.

See the [PR 1707](https://github.com/conda/conda/pull/1707) in `conda` for more details, especially on how to test this PR.

### DIFF
--- a/bin/activate.bat
+++ b/bin/activate.bat
@@ -1,69 +1,54 @@
 @echo off
+setlocal
 
-for /f "delims=" %%i in ("%~dp0..\envs") do (
-    set ANACONDA_ENVS=%%~fi
-)
+set CONDA_NEW_ENV=%1
+set CONDA_NEW_ENV=%CONDA_NEW_ENV:"=%
 
-set CONDA_NEW_NAME=%~1
-
-if "%~2" == "" goto skiptoomanyargs
-    echo ERROR: Too many arguments provided
-    goto usage
+if "%2" == "" goto skiptoomanyargs
+    (echo Error: did not expect more than one argument.) 1>&2
+    exit /b 1
 :skiptoomanyargs
 
-if not "%CONDA_NEW_NAME%" == "" goto skipmissingarg
-:usage
-    echo Usage: activate envname
-    echo.
-    echo Deactivates previously activated Conda
-    echo environment, then activates the chosen one.
+if not "%1" == "" goto skipmissingarg
+    (echo Error: no environment provided.) 1>&2
     exit /b 1
 :skipmissingarg
 
-if exist "%ANACONDA_ENVS%\%CONDA_NEW_NAME%\conda-meta" goto usenamedenv
-    for /F %%i in ("%CONDA_NEW_NAME%") do set CONDA_NEW_PATH=%%~fi
-    if exist "%CONDA_NEW_PATH%\conda-meta" goto usefullpath
-        echo No environment named "%CONDA_NEW_NAME%" exists in %ANACONDA_ENVS%
-        set CONDA_NEW_NAME=
-        set CONDA_NEW_PATH=
-        exit /b 1
-:usenamedenv
-    set CONDA_NEW_PATH=%ANACONDA_ENVS%\%CONDA_NEW_NAME%
-    goto skipmissingenv
-:usefullpath
-    for /F %%i in ("%CONDA_NEW_PATH%") do set CONDA_NEW_NAME=%%~ni
-:skipmissingenv
+if not "%1" == "--help" goto skipusage
+    (
+    echo Usage: activate ENV
+    echo.
+    echo Deactivates previously activated Conda
+    echo environment, then activates the chosen one.
+    ) 1>&2
+    exit /b 1
+:skipusage
+
+REM Use conda itself to figure things out
+REM No conda, no dice
+
+SET CONDAFOUND=
+for %%X in (conda.exe) do (set CONDAFOUND=%%~$PATH:X)
+if not defined CONDAFOUND for %%X in (conda.bat) do (set CONDAFOUND=%%~$PATH:X)
+if defined CONDAFOUND goto runcondasecretcommand
+    echo "cannot find conda to test for the environment %CONDA_NEW_ENV%"
+    exit /b 1
+
+:runcondasecretcommand
+REM Run secret conda ..checkenv command
+call "%CONDAFOUND%" ..checkenv %CONDA_NEW_ENV%
+REM EQU 0 means 0 or above on Windows ;(
+if %ERRORLEVEL% EQU 1 (
+    exit /b 1
+)
 
 REM Deactivate a previous activation if it is live
-if "%CONDA_DEFAULT_ENV%" == "" goto skipdeactivate
-    REM This search/replace removes the previous env from the path
-    echo Deactivating environment "%CONDA_DEFAULT_ENV%"...
+FOR /F "delims=" %%i IN ('"%CONDAFOUND%" ..deactivate') DO set PATH=%%i
 
-    REM Run any deactivate scripts
-    if not exist "%CONDA_DEFAULT_ENV%\etc\conda\deactivate.d" goto nodeactivate
-        pushd "%CONDA_DEFAULT_ENV%\etc\conda\deactivate.d"
-        for %%g in (*.bat) do call "%%g"
-        popd
-    :nodeactivate
+REM Activate the new environment
+FOR /F "delims=" %%i IN ('"%CONDAFOUND%" ..activate %CONDA_NEW_ENV%') DO set PATH=%%i
 
-    set CONDACTIVATE_PATH="%CONDA_DEFAULT_ENV%";"%CONDA_DEFAULT_ENV%\Scripts";"%CONDA_DEFAULT_ENV%\Library\bin"
-    call set PATH=%%PATH:%CONDACTIVATE_PATH%=%%
-    set CONDA_DEFAULT_ENV=
-    set CONDACTIVATE_PATH=
-    set PROMPT=%CONDA_OLD_PROMPT%
-    set CONDA_OLD_PROMPT=
-:skipdeactivate
+for /F %%C IN ('"%CONDAFOUND%" ..changeps1') DO set CHANGEPS1=%%C
+if "%CHANGEPS1%" == "1" set PROMPT=[%CONDA_NEW_ENV%] $P$G
 
-set CONDA_DEFAULT_ENV=%CONDA_NEW_PATH%
-echo Activating environment "%CONDA_DEFAULT_ENV%"...
-set PATH="%CONDA_DEFAULT_ENV%";"%CONDA_DEFAULT_ENV%\Scripts";"%CONDA_DEFAULT_ENV%\Library\bin";%PATH%
-set PROMPT=[%CONDA_NEW_NAME%] $P$G
-set CONDA_NEW_NAME=
-set CONDA_NEW_PATH=
-
-REM Run any activate scripts
-if not exist "%CONDA_DEFAULT_ENV%\etc\conda\activate.d" goto noactivate
-    pushd "%CONDA_DEFAULT_ENV%\etc\conda\activate.d"
-    for %%g in (*.bat) do call "%%g"
-    popd
-:noactivate
+endlocal & set PROMPT=%PROMPT%& set PATH=%PATH%& set CONDA_DEFAULT_ENV=%CONDA_NEW_ENV%
--- a/bin/deactivate.bat
+++ b/bin/deactivate.bat
@@ -1,36 +1,36 @@
 @echo off
+setlocal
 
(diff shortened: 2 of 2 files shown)


msarahan:
This looks good.  Thanks for reworking it.  Please see line comments.  In particular, I have made changes to activate.bat such that when called with no arguments, it activates the root environment.  This is desirable behavior to have consistent links to console-with-environment shortcuts that do not require activate.bat to be on PATH (see https://github.com/conda/conda-recipes/pull/439 and PRs linked therein).  What do you think about this change?  Would you be willing to adapt your code so that it behaves similarly?

bin/activate.bat
```diff
@@ -1,68 +1,54 @@
...
 
-if "%~2" == "" goto skiptoomanyargs
-    echo ERROR: Too many arguments provided
-    goto usage
+if "%2" == "" goto skiptoomanyargs
+    (echo Error: did not expect more than one argument.) 1>&2
+    exit /b 1
 :skiptoomanyargs
 
-if "%CONDA_NEW_NAME%" == "" set CONDA_NEW_NAME=%~dp0..\
```
msarahan: This part is the activating root environment part I mention in the main thread.

bin/activate.bat
```diff
@@ -1,68 +1,54 @@
...
-        set CONDA_NEW_NAME=
-        set CONDA_NEW_PATH=
-        exit /b 1
-:usenamedenv
-    set CONDA_NEW_PATH=%ANACONDA_ENVS%\%CONDA_NEW_NAME%
-    goto skipmissingenv
-:usefullpath
-    for /F %%i in ("%CONDA_NEW_PATH%") do set CONDA_NEW_NAME=%%~ni
-:skipmissingenv
+if not "%1" == "" goto skipmissingarg
```
msarahan: This section needs to go away - providing no argument should activate the root environment (deactivating anything else in the process)

bin/activate.bat
```diff
@@ -1,68 +1,54 @@
...
+    (echo Error: no environment provided.) 1>&2
+    exit /b 1
+:skipmissingarg
+
+if not "%1" == "--help" goto skipusage
+    (
+    echo Usage: activate ENV
+    echo.
+    echo Deactivates previously activated Conda
+    echo environment, then activates the chosen one.
```
msarahan: Maybe add a note here about if no ENV is added, activates root env?  If you think this is good behavior, of course (and I hope you do, it makes the shortcut stuff MUCH simpler)

bin/activate.bat
```diff
@@ -1,68 +1,54 @@
...
+    ) 1>&2
+    exit /b 1
+:skipusage
+
+REM Use conda itself to figure things out
+REM No conda, no dice
```
msarahan: This is good, but please try to make things work somehow if conda is not on PATH.  I can hard-code the path to Activate.bat into a shortcut, and I'd like to make the assumption that activate.bat can set PATH such that the rest of everything can be found.  I.e. activate.bat does not assume PATH is set up for conda, but rather can set it up based on arguments passed to activate.bat.

msarahan: If it makes sense, we can just point it to the same folder as activate.bat to look for conda.exe.  Conda can only be installed into a root environment, and should be easily locatable using a relative path given the path of activate.bat

bin/activate.bat
```diff
@@ -1,68 +1,54 @@
...
-
-    REM Run any deactivate scripts
-    if not exist "%CONDA_DEFAULT_ENV%\etc\conda\deactivate.d" goto nodeactivate
-        pushd "%CONDA_DEFAULT_ENV%\etc\conda\deactivate.d"
-        for %%g in (*.bat) do call "%%g"
-        popd
-    :nodeactivate
+FOR /F "delims=" %%i IN ('"%CONDAFOUND%" ..deactivate') DO set PATH=%%i
 
-    set CONDACTIVATE_PATH="%CONDA_DEFAULT_ENV%";"%CONDA_DEFAULT_ENV%\Scripts";"%CONDA_DEFAULT_ENV%\Library\bin"
```
msarahan: I don't see that you are keeping the \Scripts and \Library\bin folders on PATH with your new scheme.  These are pretty important to have.  Can you add them, or explain how your script keeps them?

bin/deactivate.bat
```diff
@@ -1,36 +1,36 @@
...
+    (
     echo Usage: deactivate
     echo.
     echo Deactivates previously activated Conda
     echo environment.
+    ) 1>&2
+    exit /b 1
+:skipusage
+
+if "%1" == "" goto skiptoomanyargs
```
msarahan: This has tripped up some people who try to pass env names to deactivate.  I'm a little mixed on whether it is necessary at all.  If arguments are provided, would it be OK just ignore them rather than raising an error?  Perhaps show a warning that tells them that passing env name is superfluous, since they can only deactivate the active env?

bin/deactivate.bat
```diff
@@ -1,36 +1,36 @@
...
 
+REM Use conda itself to figure things out
+REM No conda, no dice
 
-REM Deactivate a previous activation if it is live
-if "%CONDA_DEFAULT_ENV%" == "" goto skipdeactivate
-    REM This search/replace removes the previous env from the path
-    echo Deactivating environment "%CONDA_DEFAULT_ENV%"...
+SET CONDAFOUND=
+for %%X in (conda.exe) do (set CONDAFOUND=%%~$PATH:X)
```
msarahan: Same note as above: please do not assume conda is on PATH, but rather try to find it using relative paths and knowledge of deactivate.bat path.

bin/deactivate.bat
```diff
@@ -1,36 +1,36 @@
...
-    if not exist "%CONDA_DEFAULT_ENV%\etc\conda\deactivate.d" goto nodeactivate
-        pushd "%CONDA_DEFAULT_ENV%\etc\conda\deactivate.d"
-        for %%g in (*.bat) do call "%%g"
-        popd
-    :nodeactivate
+:runcondasecretcommand
+REM activate conda root environment
+FOR /F "delims=" %%i IN ('conda ..activateroot') DO set PATH=%%i
 
-    set CONDACTIVATE_PATH="%CONDA_DEFAULT_ENV%";"%CONDA_DEFAULT_ENV%\Scripts";"%CONDA_DEFAULT_ENV%\Library\bin";
```
msarahan: Same note as above: please keep \Scripts and \Library\bin

msarahan:
@malev waiting on revision from @wulmer on this one.  Not going to be part of immediate release.

wulmer:
@msarahan Thanks for the fast answer. Sorry that mine took a bit longer.

Just let me recap your remarks:

1. `activate.bat` should activate root when no argument is given + usage documentation and log message should mention that.
2. Should work with `conda.exe` not being on the `%PATH%` (under the assumption that `conda.exe` is just at `%~dp0conda.exe`)
3. Same for `deactivate.bat`: allow argument but ignore with message, work without `conda.exe` being on the `%PATH%`.

Here my answers/remarks:

My PR tries to unify Windows and Linux/OS X behaviour. It does this by using internal conda mechanisms like `conda ..deactivate` or `conda ..activate` to modify the `%PATH%` variable. As the internal mechanisms work mostly similar for Windows and Linux now, it is a bit ugly to break it again and make the Windows part behave differently. One of the great things that this PR fixes is that it actually *removes* the root environment from the PATH when another environment is activated. This makes command line behaviour much more consistent, e.g. if you have no IPython in your environment, running `python` is from your env, running `ipython` comes from root env !?!
So I am not sure how well this plays with point 2.) but I will try.

1. Making `activate` activate the root environment should be no problem. However I it breaks many of the tests in `conda/tests/test_activate.py` as the behaviour on Linux/OS X is different (AFAIK, at least it was in 2.1.0). As I am using both Windows and Linux, I don't like the idea of having different syntaxes on these platforms. What about changing the Linux behaviour also? This would keep the internal behaviour and the tests consistent between the platforms?
2. I also don't think this is a problem. As you wrote `conda` can easily be found under `%~dp0`. 
3. Same as 1. This leads to an inconsistency between Windows and Linux/OS X. What would be a reasonable common behaviour?

Would you mind adding your activate specifications to the tests in `conda/tests/test_activate.py`? I can't find any recent tests about the new activate behaviour and not introducing regressions is really hard without testing. It would help me a lot if I had some tests I could code against.
BTW: I noticed that my [original test process description in PR 1707](https://github.com/conda/conda/pull/1707#issue-111506545) was wrong. Just running `python setup.py develop` for `conda-env` doesn't take care that the correct `(de)activate.bat` scripts are executed. I haven't found a proper solution for that. Currently, I have to copy my modifications after every change directly into the Miniconda `Scripts/` directory. Not very satisfying.

One word about how paths are added/removed to/from `%PATH%`: The calculation of the paths is done by `conda` itself. The lines with `FOR /F "delims=" %%i IN ('"%CONDAFOUND%" ..activate %CONDA_NEW_ENV%') DO set PATH=%%i` and `FOR /F "delims=" %%i IN ('"%CONDAFOUND%" ..deactivate') DO set PATH=%%i` do all the path adjustment. This is much easier than doing path calculations in Batch files. Currently, the changes in conda/conda#1707 only add the `Scripts` and the base dir to `%PATH%`.

msarahan:
Thanks for this detailed and useful response.  This is a fundamental change
in operation philosophy.  I am pretty sure that it is for the better, but I
need to make sure we don't break the existing Windows ecosystem.

Tests I think should pass (that I need to write):
Assumptions / setup:
1. environment "test_main" exists at <conda install dir>\envs\test_main
2. environment "test_alt_location" exists at C:\envs\test_alt_location (or
whatever)
3. Environment variable CONDA_ENV_ is set to C:\envs;<conda install
dir>\envs
4. activate.bat is on PATH (this is default.  Some tests below specify
otherwise.)

Tests
- from cmd.exe, user can call activate.bat (or .ps1?)
   - "test_main" activated correctly (by name)
   - "test_alt_location" activated correctly (by name)
- set CONDA_ENV to not include <conda install dir>
   - verify that "test_main" is NOT found by activate
- given complete path of SOMETHING (not necessarily activate.bat), one can
activate any environment (including root.)  This is used for creating
shortcuts.  The SOMETHING need not be activate.bat.  It can be any helper
script, if it makes implementation simpler or more consistent across
platforms.
    - set PATH such that activate.bat and conda.exe are not on PATH.
    - provide full path to activate.bat or helper bat, plus environment by
name; verify that shell with appropriate env variables for env is active
    - provide full path to activate.bat or helper bat, plus environment by
full path spec; verify that shell with appropriate env variables for env is
active
- Activate a path given a relative path to an environment.  From C:\,
"activate envs\test_main"
   - check handling of forward and backslashes
   - check handling of spaces in path
   - specify relative paths both forward from cwd, and backward
(..\blah\env)
- Activate a path given an absolute path to an environment
   - check handling of forward and backslashes
   - check handling of spaces in path

I think that covers it, but please add anything I might have missed.  This
kind of thing will probably mostly involve a lot of subprocess calls, which
I hope examples already exist, but might not.  I'm buried with the Anaconda
release right now and can't give this a proper technical evaluation and
test-writing time, but I don't want this to get lost again.  If you have
time to write tests, please feel free to add them here.  Otherwise, I'll
get to them soon.

I think on the whole, we should aim to have Linux/OSX stay as it is, and
change windows behavior.  I say this because I think more people probably
use activate on the command line in Linux and OSX.  If there's compelling
reason to reconsider that (if, for example, running activate with no args
makes life on Linux better), I'll raise it with the Continuum team.

I'm going to put this at the top of my list to resolve when the Anaconda
release is complete.  If you don't hear from me by Halloween, please howl,
shriek, or express your choice of Halloween-flavored (virtual) vocalization.

Very cool about just using Conda itself to adjust the path.  I need to
study that and get a better handle on it, but that is MUCH nicer.  I am
also eager to discuss how this affects the many requests for proper
powershell support.  Does this make it easier?  Does it change it at all?

msarahan:
Integrating this at I have created PRs: https://github.com/conda/conda/pull/1727 and https://github.com/conda/conda-env/pull/189 because I don't know how to issue changes to your PR.  If you make any further changes, let me know and I will integrate them.  If you want to teach me how to fish (figuratively) - I'd be OK to learn how to contribute these changes directly to your PR rather than starting new ones.

masaeedu:
@msarahan I think you can fork @wulmer's fork, make some changes, then PR those to the branch he is PR-ing here. Once he merges the changes in, they will be reflected in this PR.

msarahan:
Superseded by https://github.com/conda/conda/pull/1727.  Closing.

jphilippou27:
I'm having some trouble getting the activate command to work when changing the envs directories variable. I posted about it in more detail [here](http://stackoverflow.com/questions/41248328/conda-activate-env-not-changing-file-directory). Please take a look!

PR #183, opened 2015-10-17
Remove calls to $_THIS_DIR/conda to make activate faster

gqmelo:
This is an attempt to make the activate script faster. I don't think this PR is ready to merge (or if it ever will), but I would like to get opinions about this and if possible to contribute this changes to official ```conda-env``` release.

Removing calls to conda command makes activate about **ten times faster**.
This may seem irrelevant for interactive shell sessions, but if you want for example to make your IDE
 run commands (python, gcc, etc.) on different environments it is very important for the activate to be fast.

I don't know if I covered all the cases, but I was able to make it work with the activate tests present on ```conda``` repository (with a [minor modification] (https://github.com/gqmelo/conda/tree/feature/faster-activate))

### DIFF
--- a/bin/activate
+++ b/bin/activate
@@ -23,14 +23,34 @@ fi
 
 _THIS_DIR=$(dirname "$_SCRIPT_LOCATION")
 
+echoerr() {
+    echo "$@" 1>&2
+}
+
+# Check arguments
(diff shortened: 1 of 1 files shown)


malev:
@gqmelo thanks for contributing. I'm working on a new release for conda-env so I won't have time to take a deep look. Now

kalefranz:

Issue #181, opened 2015-10-14
Bash "set -u" exits with error when activating env


SlaterByte:
$CONDA_OLD_PS1 is not set when activating outside of any conda env. When included in a Bash script with set -u, script will blow up with Unbounded Variable error.

Link to line:
https://github.com/conda/conda-env/blob/develop/bin/activate#L51

Not sure the best way to test if variable exists, then test if it is not null. Maybe an early test if variable exists and then if it does not set it to null.

malev:
good advice. Feel free to open a PR if you want

SlaterByte:
Do I need to provide anymore input?

PR #184, opened 2015-10-19
Set potentially unset variables to null. 

SlaterByte:
This satisfies issue #181 and has been tested with both bash and zsh with 'set -ue' and with 'set -o pipeline' with bash.

### DIFF
--- a/bin/activate
+++ b/bin/activate
@@ -3,7 +3,9 @@
 # Ensure that this script is sourced, not executed
 # Also note that errors are ignored as `activate foo` doesn't generate a bad
 # value for $0 which would cause errors.
-if [[ -n $BASH_VERSION ]] && [[ "$(basename "$0" 2> /dev/null)" == "activate" ]]; then
+# Also note, variables that could potentially be unset should have parameter
+# expansion set them to null. ex: ${VARIABLE:=''}
+if [[ -n ${BASH_VERSION:=''} ]] && [[ "$(basename "$0" 2> /dev/null)" == "activate" ]]; then
     >&2 echo "Error: activate must be sourced. Run 'source activate envname'
 instead of 'activate envname'.
 "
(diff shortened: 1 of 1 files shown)


ahmadia:
I just reviewed this.  It seems that in every case you prefer the `:=` form of parameter expansion (sets the variable) as opposed to `:` (tests the variable, but doesn't modify it).  Don't we want the latter in most of these cases?

bin/activate
```diff
@@ -3,7 +3,9 @@
 # Ensure that this script is sourced, not executed
 # Also note that errors are ignored as `activate foo` doesn't generate a bad
 # value for $0 which would cause errors.
-if [[ -n $BASH_VERSION ]] && [[ "$(basename "$0" 2> /dev/null)" == "activate" ]]; then
+# Also note, variables that could potentially be unset should have parameter
+# expansion set them to null. ex: ${VARIABLE:=''}
+if [[ -n ${BASH_VERSION:=''} ]] && [[ "$(basename "$0" 2> /dev/null)" == "activate" ]]; then
```
ahmadia: There's no reason to set BASH_VERSION in this test.

bin/activate
```diff
@@ -12,9 +14,9 @@ instead of 'activate envname'.
 fi
 
 # Determine the directory containing this script
-if [[ -n $BASH_VERSION ]]; then
+if [[ -n ${BASH_VERSION:=''} ]]; then
```
ahmadia: Same here.  I think you want: `${BASH_VERSION:''}` instead.

ahmadia:
I like the idea of this.  I think the preferred form that should be safe in both Bash and Zsh is `${VAR:-''}`.

ahmadia:
In this situation `${VAR-''}` should be morally equivalent.

ahmadia:
@SlaterByte - it would be great if we could get this fix in.  Let me know if you'd like to try a second pass at this.

kalefranz:

PR #187, opened 2015-10-21
activate.bat: removed incorrect quotes

ukoethe:
Since `%CONDA_ENVS_PATH%` is enclosed within quotes everywhere, the present `set` command causes the path to be quoted twice, which doesn't work.

### DIFF
--- a/bin/activate.bat
+++ b/bin/activate.bat
@@ -1,8 +1,8 @@
 @echo off
-REM Check for CONDA_EVS_PATH environment variable
+REM Check for CONDA_ENVS_PATH environment variable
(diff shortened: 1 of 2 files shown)


msarahan:
Thanks hugely for this fix.  Please see https://github.com/conda/conda-env/pull/188 - I would greatly appreciate your review.

msarahan:
I am going to merge this now, with the understanding that it will be completely overwritten by https://github.com/conda/conda-env/pull/189 and https://github.com/conda/conda/pull/1727

If anyone is using conda-env from source, I hope this helps you.

We have removed packages for conda-env 2.4.3 from the repository, so we hope that people will not have further issue going forward.

PR #189, opened 2015-10-22
Wulmer fix/activate behaviour on windows

msarahan:
This integrates https://github.com/conda/conda-env/pull/182 along with some minor changes:

- calling activate with no arguments activates the root environment
- Conda.bat is found relative to the activate.bat script.  For all but root env, this conda.bat script will call the root conda.bat script, which should be created by conda itself in each environment as it is created.

Remaining work:
- deactivate.bat needs to use the relative conda.bat logic similarly to activate.


bin/activate.bat
```diff
@@ -1,78 +1,44 @@
...
-REM First match will be the one used
+REM Use conda itself to figure things out
 
-for %%F in ("%CONDA_ENVS_PATH:;=" "%") do (
-    if exist "%%~F\%CONDA_NEW_NAME%\conda-meta" (
-       set CONDA_NEW_PATH=%%~F\%CONDA_NEW_NAME%
-       goto found_env
-    )
-)
+SET "CONDA_EXE=%~dp0\conda.bat"
```
ukoethe: This is wrong -- the command is `conda.exe`

ukoethe:
I tested this PR, along with https://github.com/conda/conda/pull/1727, on Windows, and it does __not__ work. I indicated some errors in the code, but there are more problems. Unfortunately, I don't have time for more detailed comments right now. Please do __not__ merge before fixing this.

msarahan:
Thanks for testing.  I am also running into issues.  Rest assured, this will not be merged before it is working.

msarahan:
PS: your stamp of approval would give me greater confidence in saying "it works" - I will ping you for review, if that's OK.

bin/activate.bat
msarahan: They both exist.  I think the key difference is that conda.bat might actually just contain code pointing to the root conda.bat, which then knows about conda.exe.  Conda can only exist at the root level, or else the concept of the root level is lost, and deactivating to some reference state is more ambiguous.

ukoethe: I double-checked -- there is no `conda.bat` in my installation (or it is hiding really well :-).

msarahan: Are you also using the development version of conda from conda/conda#1727 ? Note that you may need to manually delete/move/rename/whatever the <prefix>/Lib/site-packages/conda folder (and the conda-env folder there, too), or else the conda-installed packages take preference.  This has bitten me quite hard in debugging these packages.

ukoethe: I did that, but still no `conda.bat`. How am I supposed to install it?

ukoethe: Oops, now I suddenly have `conda.bat` in one of my environments - it was apparently silently installed by some other `conda install` command.

bin/activate.bat
```diff
@@ -1,78 +1,44 @@
...
+FOR /F "delims=" %%i IN ('"%CONDA_EXE%" ..activate %CONDA_NEW_ENV%') DO set PATH=%%i
+if %ERRORLEVEL% GTR 0 (
+exit /b 1
+)
+for /F %%C IN ('"%CONDA_EXE%" ..changeps1') DO set CHANGEPS1=%%C
+if "%CHANGEPS1%" == "1" set PROMPT=[%CONDA_NEW_ENV%] $P$G
+if %ERRORLEVEL% GTR 0 (
+exit /b 1
+)
+endlocal & set PROMPT=%PROMPT%& set PATH=%PATH%& set CONDA_DEFAULT_ENV=%CONDA_NEW_ENV%
```
ukoethe: The variable name `CONDA_DEFAULT_ENV` is now inconsistent with the code in https://github.com/conda/conda/pull/1727, where you changed it into `CONDA_ACTIVE_ENV`. As I said elsewhere, I vote for the new name.

kalefranz:

PR #195, opened 2015-10-29
Implement --prune options to "conda env update"

### LINKED ISSUE #190: Conda env update doesn't remove installed packages
Hi,

Suppose an environment created with the following `environment.yml`:

```yml
name: conda-env-test
dependencies:
  - python
  - py
```

If the user removes `py` from the dependencies list in the file and executes `conda env update` on this environment, `conda-env` won't remove `py` from the environment.

Is this behavior a bug? If so, I would be happy to work on a PR to fix this.

nicoddemus:
Fix #190 

I could not figure out how to write a proper test for this, any help here is appreciated.

This PR depends on conda/conda#1754

### DIFF
--- a/conda_env/cli/main_update.py
+++ b/conda_env/cli/main_update.py
@@ -47,6 +47,12 @@ def configure_parser(sub_parsers):
         default='environment.yml',
     )
     p.add_argument(
+        '--prune',
+        action='store_true',
+        default=False,
+        help='remove installed packages not defined in environment.yml',
+    )
+    p.add_argument(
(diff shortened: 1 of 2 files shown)


malev:
Looking good in this side. Let's wait for the conda team to respond on the other

ushuz:
Looking forward to this.

mwiebe:
Also ping @kalefranz and @msarahan on this one. This together with https://github.com/conda/conda-env/pull/229 would help our use-case a lot.

PR #201, opened 2015-11-10
`--quiet` arg action should store `True`

faph:
The `--quiet` argument doesn't seem to work. I think it should store `True` as it defaults to `False`.


malev:
It's working great! Thanks @faph

PR #203, opened 2015-11-18
Requirements spec

malev:
A replacement for #172
@dan-blanchard, what do you think?


dan-blanchard:
This is missing a big piece of my PR, which is the ability to have conda figure out which dependencies are available via conda and which it needs to use pip for.  That's actually the whole reason I put together my PR.  This is just an alternative format to list conda dependencies.

malev:
oh, so you want to have a single list of commands and having conda-env figure out which one is in conda and which on is in pypi? am I right?

dan-blanchard:
Yeah, exactly.  This is mostly useful because most of the projects I work on the other devs do not use conda, so having to always go through the requirements list and then manually conda install everything myself was a pain.  Also, dependencies are the real issue here, because often times something won't be available via conda, but it relies on numpy or pandas, which should be installed via conda.

That's why my PR has [this function](https://github.com/dan-blanchard/conda-env/blob/feature/pip_requirements.txt/conda_env/env.py#L75-L104) to get all of the pip dependencies.  And then the logic [here](https://github.com/dan-blanchard/conda-env/blob/feature/pip_requirements.txt/conda_env/env.py#L113-L129) that checks what's available via conda and what is not (and it knows to skip editable packages, because those are only for pip).

dan-blanchard:
Also, pip needs to be called like [this](https://github.com/dan-blanchard/conda-env/blob/feature/pip_requirements.txt/conda_env/installers/pip.py#L12-L18) in order for editable packages to be checked out within the conda environment instead of just in the current directory.  And passing on the command-line arguments for editable packages is tricky too.

conda_env/specs/requirements.py
```diff
@@ -0,0 +1,43 @@
...
+    and returns an Environment object from it.
+    '''
+    msg = None
+
+    def __init__(self, filename=None, name=None, **kwargs):
+        self.filename = filename
+        self.name = name
+        self.msg = None
+
+    def can_handle(self):
```
goanpeca: It would be nice to have normal methods after private ones...

so

```python
def __init__  ...

def _valid_file ...

def _valid_name ....

def can_handle ...
```

conda_env/specs/requirements.py
```diff
@@ -0,0 +1,43 @@
+import os
+from .. import env
+
```
goanpeca: It is nice to have imports separated by the origin like...


```python
# Standard library imports
import os

# Local Imports
from .. import env
```

tests/specs/test_requirements.py
```diff
@@ -0,0 +1,26 @@
+import unittest
+from conda_env import env
+from conda_env.specs.requirements import RequirementsSpec
+from .. import utils
+
```
goanpeca: Same as comment from before :-)

malev:
1) Gonna merge this now.
2) Gonna work on your PR, I think it's gonna be a good feature.
3) Thanks for the tip on pip

conda_env/specs/requirements.py
```diff
@@ -0,0 +1,44 @@
...
+    def _valid_file(self):
+        if os.path.exists(self.filename):
+            return True
+        else:
+            self.msg = "There is no requirements.txt"
+            return False
+
+    def _valid_name(self):
+        if self.name is None:
+            self.msg = "Environment with requierements.txt file needs a name"
```
wadawson, line 27: There is a typo here:
requierements.txt should be requirements.txt

PR #205, opened 2015-11-19
Jinja2 environment files

malev:
Proof of concept.
Will cover part of #88
Note: still running tests. 
@campos-ddc and @nicoddemus what do you guys think?
I still like the idea of the environment variables, but I have left that out of this PR.

### DIFF
--- a/conda_env/specs/__init__.py
+++ b/conda_env/specs/__init__.py
@@ -2,13 +2,15 @@ from .binstar import BinstarSpec
 from .yaml_file import YamlFileSpec
 from .notebook import NotebookSpec
 from .requirements import RequirementsSpec
+from .jinja2_file import Jinja2Spec
 from ..exceptions import SpecNotFound
 
 all_specs = [
     BinstarSpec,
     NotebookSpec,
     YamlFileSpec,
-    RequirementsSpec
+    RequirementsSpec,
+    Jinja2Spec
 ]
 
 
--- a/conda_env/specs/jinja2_file.py
+++ b/conda_env/specs/jinja2_file.py
@@ -1,0 +1,36 @@
+try:
+    import jinja2
+except ImportError:
+    jinja2 = None
+from .. import env
+from ..exceptions import EnvironmentFileNotFound, Jinja2NotInstalled
+
+
+class Jinja2Spec(object):
+    _environment = None
+
+    def __init__(self, filename=None, **kwargs):
+        self.filename = filename
+        self.msg = None
+
+    def can_handle(self):
+        try:
+            self._environment = env.from_yaml(
+                self._jinja2(self.filename)
+            )
+            return True
+        except (EnvironmentFileNotFound, Jinja2NotInstalled, IOError) as e:
+            self.msg = str(e)
+            return False
+
+    def _jinja2(self, filename):
+        if jinja2 is None:
+            raise Jinja2NotInstalled()
+        with open(filename) as jinjafile:
+            return jinja2.Template(jinjafile.read()).render()
+
+    @property
+    def environment(self):
+        if not self._environment:
+            self.can_handle()
+        return self._environment
--- a/tests/specs/test_jinja2_file.py
+++ b/tests/specs/test_jinja2_file.py
@@ -1,0 +1,20 @@
+import unittest
+
+from conda_env import env
+from conda_env.specs.jinja2_file import Jinja2Spec
+
+from .. import utils
+
+
+class TestJinja2(unittest.TestCase):
+    def test_no_environment_file(self):
+        spec = Jinja2Spec(name=None, filename='not-a-file')
+        self.assertEqual(spec.can_handle(), False)
+
+    def test_no_name(self):
+        spec = Jinja2Spec(filename=utils.support_file('jinja2.yml'))
+        self.assertEqual(spec.can_handle(), True)
+
+    def test_get_environment(self):
+        spec = Jinja2Spec(filename=utils.support_file('jinja2.yml'))
+        self.assertIsInstance(spec.environment, env.Environment)
--- a/conda_env/exceptions.py
+++ b/conda_env/exceptions.py
@@ -63,3 +63,11 @@ class NBFormatNotInstalled(CondaEnvRuntimeError):
         conda install nbformat
         """
         super(NBFormatNotInstalled, self).__init__(msg)
+
+
+class Jinja2NotInstalled(CondaEnvRuntimeError):
+    def __init__(self):
+        msg = """jinja2 is not installed. Install it with:
+        conda install jinja2
+        """
+        super(NBFormatNotInstalled, self).__init__(msg)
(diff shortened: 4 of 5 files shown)


conda_env/specs/jinja2_file.py
```diff
@@ -0,0 +1,36 @@
+try:
+    import jinja2
+except ImportError:
+    jinja2 = None
+from .. import env
+from ..exceptions import EnvironmentFileNotFound, Jinja2NotInstalled
+
+
+class Jinja2Spec(object):
+    _environment = None
```
campos-ddc: Why is this a class variable instead of instance?

conda_env/specs/jinja2_file.py
```diff
@@ -0,0 +1,36 @@
...
+
+    def __init__(self, filename=None, **kwargs):
+        self.filename = filename
+        self.msg = None
+
+    def can_handle(self):
```
campos-ddc: Calling `can_handle` multiple times will load the file multiple times, I don't think it really matters, but just letting you know :)

campos-ddc:
I think this works just fine, I should have separated jinja from other features the first time!

conda_env/specs/jinja2_file.py
```diff
@@ -0,0 +1,36 @@
...
+            return True
+        except (EnvironmentFileNotFound, Jinja2NotInstalled, IOError) as e:
+            self.msg = str(e)
+            return False
+
+    def _jinja2(self, filename):
+        if jinja2 is None:
+            raise Jinja2NotInstalled()
+        with open(filename) as jinjafile:
+            return jinja2.Template(jinjafile.read()).render()
```
nicoddemus: We probably want to add some variables to the namespace evaluation for it to be useful, like `os` and `sys` modules.

conda_env/specs/jinja2_file.py
malev: mmh you are right about that, maybe we can think on some sort of proxy. That would be for another PRI guess

conda_env/specs/jinja2_file.py
malev: can we add a module to the context? can't we import 'em from the template? I'm gonna research how is `conda` doing this.

conda_env/specs/jinja2_file.py
malev: because https://github.com/conda/conda-env/blob/jinja2/conda_env/specs/jinja2_file.py#L34
I want to be sure it's been defined already

campos-ddc: I see.

Normally I'd just put that in `__init__`, defining as `None`.
Leaving it as a class variable will get you into trouble if this code is ran multiple times in the same process (certainly will happen if there are multiple tests for this)

conda_env/specs/__init__.py
```diff
@@ -2,13 +2,15 @@
...
+from .jinja2_file import Jinja2Spec
 from ..exceptions import SpecNotFound
 
 all_specs = [
     BinstarSpec,
     NotebookSpec,
     YamlFileSpec,
-    RequirementsSpec
+    RequirementsSpec,
+    Jinja2Spec
```
nicoddemus: How is this supposed to be used from an user's point of view? I mean, if `Jinja2Spec` is the last one in this list, won't `YamlFileSpec` be end up being picked up instead every time?

tests/specs/test_jinja2_file.py
```diff
@@ -0,0 +1,20 @@
...
+        spec = Jinja2Spec(name=None, filename='not-a-file')
+        self.assertEqual(spec.can_handle(), False)
+
+    def test_no_name(self):
+        spec = Jinja2Spec(filename=utils.support_file('jinja2.yml'))
+        self.assertEqual(spec.can_handle(), True)
+
+    def test_get_environment(self):
+        spec = Jinja2Spec(filename=utils.support_file('jinja2.yml'))
+        self.assertIsInstance(spec.environment, env.Environment)
```
nicoddemus: I think it would be interesting here to test which dependencies were loaded, and make `jinja2.yml` install different dependencies based on some variable that we may artificially put in `sys` or `os`.

Something like this for example:

```yml
dependencies:
  - flask
  {% if sys.conda_env_testing == '1' %}
  - jinja2
  {% else %}
  - yaml
  {% endif %}
```

```python
def test_get_environment(self):
    sys.conda_env_testing = '1'
    self.addCleanup(delattr, sys, 'conda_env_testing')
    spec = Jinja2Spec(filename=utils.support_file('jinja2.yml'))
    self.assertIsInstance(spec.environment, env.Environment)
    self.assertEqual(spec.dependencies['conda'], ['flask', 'jinja2'])

    sys.conda_env_testing = '2'
    spec = Jinja2Spec(filename=utils.support_file('jinja2.yml'))
    self.assertEqual(spec.dependencies['conda'], ['flask', 'yaml'])
```

conda_env/specs/__init__.py
tadeu: If the file is pure yaml, yes, but if it has jinja markers, `YamlFileSpec.can_handle` should return `False`.

nicoddemus: Oh sure, thanks. :smile:

tadeu: Actually, perhaps it is possible for a file to have jinja2 delimiters and still be a valid yaml file! In this case, your concern is still valid.

malev: @tadeu do you have an example of that?

tadeu: Here is an example:

```yaml
name: test
dependencies:
- python={{ os.environ['PYTHON_VER'] }}
- six=1.10.0
```

It is not malformed yaml and will be parsed right, it will only fail at dependency resolve time.

malev: @tadeu you are right about that

anjos:
This looks like a promising patch, but nothing really happened since nov 2015. Is this, by any chance, going forward?

nicoddemus:
@anjos this has not moved forward in `conda-env` unfortunately.

We plan to release [conda-devenv](https://github.com/ESSS/conda-devenv) in the next couple of months. It includes Jinja2 support and a few other features like includes and setting environment variables. It has been used internally by us for some time now, it is not `1.0` yet because we are still doing some tests and it needs better docs.

anjos:
@nicoddemus thanks for the link, looks like a promising project. Is installation info on RTD accurate? I can't seem to find neither a PyPI package or conda-forge feedstock available.

nicoddemus:
It is not accurate, it is still under heavy development (but active development, I'm working on it right now 😉). It will probably be available as a package on [conda-forge](https://github.com/conda-forge/).

nicoddemus:
@anjos we have a recipe in place already: conda-forge/staged-recipes#2657, should be available for use soon. 👍

nicoddemus:
Available at https://github.com/conda-forge/conda-devenv-feedstock
Docs: http://conda-devenv.readthedocs.io/en/latest/

PR #212, opened 2015-12-07
Activate keep script dir on PATH

msarahan:
This addresses problems reported by several users about conda breaking after activating an environment, due to loss of activate scripts on PATH.  It does so by appending the root Scripts dir onto PATH during any deactivation.  This should always be at the end of PATH, so it hopefully will not interfere.  This does not require modification of the system or user PATH setting, but it is only persistent in a given terminal session.  

This fix also addresses issues with spaces in user paths and environment names.

Ping @groutr @ccordoba12

### DIFF
--- a/bin/activate.bat
+++ b/bin/activate.bat
@@ -68,8 +68,6 @@ IF "%CONDA_DEFAULT_ENV%" == "" GOTO skipdeactivate
 
     SET "CONDACTIVATE_PATH=%CONDA_DEFAULT_ENV%;%CONDA_DEFAULT_ENV%\Scripts;%CONDA_DEFAULT_ENV%\Library\bin"
     CALL SET "PATH=%%PATH:%CONDACTIVATE_PATH%=%%"
-    REM Trim trailing semicolon, if any
-    IF "%PATH:~-1%"==";" SET "PATH=%PATH:~0,-1%"
     SET CONDA_DEFAULT_ENV=
 :skipdeactivate
 
@@ -94,6 +92,12 @@ IF NOT EXIST "%CONDA_DEFAULT_ENV%\etc\conda\activate.d" GOTO noactivate
     FOR %%g IN (*.bat) DO CALL "%%g"
     POPD
 :noactivate
+
+REM Trim trailing semicolon, if any
+IF "%PATH:~-1%"==";" SET "PATH=%PATH:~0,-1%"
+
+REM Clean up any double colons we may have ended up with
+SET "PATH=%PATH:;;=;%"
 
 ENDLOCAL & (
     SET "PATH=%PATH%"
--- a/bin/activate.bat
+++ b/bin/activate.bat
@@ -7,6 +7,9 @@ IF "%CONDA_ENVS_PATH%" == "" (
     CALL :NORMALIZEPATH CONDA_ENVS_PATH "%~dp0..\envs"
 )
 
+REM Used for deactivate, to make sure we restore original state after deactivation
+IF "%CONDA_PATH_BACKUP%" == "" (SET "CONDA_PATH_BACKUP=%PATH%")
+
 set "CONDA_NEW_NAME=%~1"
 
 IF "%~2" == "" GOTO skiptoomanyargs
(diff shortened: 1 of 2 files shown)


ilanschnell:
Is this change tested?  Should we merge it and add a new conda-env release to Anaconda 2.4.1?

msarahan:
I have tested this.  I would appreciate if at least one other person tested this before merging.

The most challenging test of this is to create an installation that has a space in it.  I have tested this with installations with spaces, as well as env names with spaces.

ccordoba12:
I'm testing it right now.

msarahan:
Thanks Carlos.  Please make sure that any env you activate is actually active.  I like to use a root env of one python version, then create envs of another python version.  I then run python after activating.  I have had instances where the prompt was changed, and the PATH was modified, but I still got the wrong Python somehow.  I think I have that worked out, but I worry about it.

ccordoba12:
My findings:

1. `activate foo` is working as expected! I mean, it doesn't remove the root env from PATH and it works for paths with spaces on it.
2. `deactivate` is failing with this PR. After running it, two dirs are removed from PATH:

    * `<path to root env>\Library\bin`
    * `<path to root env>`

    The only one that is preserved is

    * `<path to root env>\Scripts`

    This means that you can't run `python` after `deactivate` (that's how I noticed it).

msarahan:
So, that deactivate behavior is actually by design.  Deactivating is intended to take things off PATH, not restore the root environment.  To activate root, you can just type ```activate``` with no argument.

This PR is meant to ensure that people are always able to call activate and deactivate - not necessarily Python.

ccordoba12:
But that's not how things have worked so far. `deactivate` just removes from PATH the new entries added by `activate`. So you come back to the root env if you activated just one env.

Questions:
1. When did this change?
2. Is this going to be the default behavior on Linux and Mac too?

msarahan:
I'm not sure that's true.  What has been more true in the past has been that the root environment is most often added to PATH (either user or system).  If deactivate is removing stuff that is in the environment variable, then that is a problem.  Otherwise, it is removing stuff that is the result of an implicit activation (starting the Anaconda prompt, for example).

Also, this is not a stack: Deactivate will always take you back to whatever state you have in your base environment variables.  I don't think that has changed, nor is it going to change.

ccordoba12:
> If deactivate is removing stuff that is in the environment variable, then that is a problem

Yes, that's what's doing. During installation, I decided to add Anaconda2 to my PATH. And after `deactivate` some of its values (as I mentioned above) are removed, but not permanently just for the current session.

> Deactivate will always take you back to whatever state you have in your base environment variables

This is what I'm seeing:

seleccion_001

So I'd say `deactivate` is not taking me to my initial state. That's why I think this is a bug :-)

msarahan:
Cool, agreed.  One quick question, though: it is desirable to have the root env hidden when running in an env, right?  I think people have requested this for the sake of not "accidentally" picking up other versions.  I just need to make sure that things are restored after deactivation.

ccordoba12:
> it is desirable to have the root env hidden when running in an env, right?

If by *hidden*, you mean any of these options:

1. After the paths of the current env
2. At the end of PATH
3. Just leaving `<root env>\Scripts` in PATH (so that `conda` and `activate/deactivate` work)

then I agree :-)

msarahan:
Yes, I mean option 3 here.  Hide PREFIX and Library/bin, but make sure they are replaced (if they were hidden) upon deactivation.

msarahan:
1 and 2 are not good options, because stuff can still get picked up in strange, unpredictable ways.

ccordoba12:
Yep, 3. is the safest option.

ccordoba12:
Even better would be to install `conda` and `activate/deactivate` in its own prefix, so we can add only them to PATH.

msarahan:
Yeah, that's what my other PR does.  That needs more discussion, I think.

msarahan:
@ccordoba12 one more test, please.  I think I have it fixed.  I just store a backup of the original PATH when activate is first used.

ccordoba12:
Like in the screenshot I showed you, right?

ccordoba12:
The problem seems to be solved! Great work!

seleccion_002

msarahan:
This is ready.  Merging.

ilanschnell:
Great!

PR #213, opened 2015-12-10
feature/preprocessing selectors

Korijn:
This feature allows you to filter dependencies based on preprocessing selectors, similar to [how they work in conda-build](http://conda.pydata.org/docs/building/meta-yaml.html#preprocessing-selectors). You can supply custom selectors to the `create` and `update` commands, like so:

```yaml
name: selectors
selectors:
  - test
  - docs
dependencies:
  - python
  - pytest  # [test]
  - sphinx  # [docs]
  - doctest  # [test and docs]
```

* `conda env create` will install only `python`
* `conda env create --select test` will install `python` and `pytest`
* `conda env create --select test --select docs` will install everything
* `conda env create --select all` will install everything

Unit test included.

This also adds platform selectors, e.g.:

```yaml
name: platforms
dependencies:
  - python
  - pywin32  # [win]
```

The available platform selectors are:

* `linux`
* `linux32`
* `linux64`
* `osx`
* `win`
* `win32`
* `win64`
* `unix` (is equivalent to `osx or linux`)
* `unix32`
* `unix64`
* `x64`
* `x86`

You can specify arbitrary python expressions, since they are `eval`d, and you can use the `os` and `os.environ` (aliased as `environ`) objects as well, e.g.:

```yaml
name: platforms
dependencies:
  - python
  - pywin32  # [win32 and win64]
  - sphinx_rtd_theme  # [not (environ.get('READTHEDOCS', None) == 'True')]
```

Please note that most of the design choices were made so that the functionality mirrors that of [conda-build's preprocessing selectors](http://conda.pydata.org/docs/building/meta-yaml.html#preprocessing-selectors), so that users don't need to become familiar with different syntax.

### DIFF
--- a/conda_env/env.py
+++ b/conda_env/env.py
@@ -4,7 +4,6 @@ from copy import copy
 import os
 import re
 import sys
-import platform
 
 # TODO This should never have to import from conda.cli
 from conda.cli import common
@@ -80,23 +79,23 @@ def from_file(filename, selectors=None):
 
 
 def ns_cfg(custom_selectors, selectors=None):
-    plat = sys.platform
-    arch = platform.architecture()
+    plat = sys.platform  # see https://docs.python.org/2/library/sys.html#sys.platform
+    is_x64 = sys.maxsize > 2**32  # from https://docs.python.org/2/library/platform.html#cross-platform
     d = dict(
         linux=plat.startswith('linux'),
-        linux32=plat.startswith('linux') and '32' in arch,
-        linux64=plat.startswith('linux') and '64' in arch,
+        linux32=plat.startswith('linux') and not is_x64,
+        linux64=plat.startswith('linux') and is_x64,
         osx=plat.startswith('darwin'),
         win=plat.startswith('win32'),
-        win32=plat.startswith('win32') and '32' in arch,
-        win64=plat.startswith('win32') and '64' in arch,
+        win32=plat.startswith('win32') and not is_x64,
+        win64=plat.startswith('win32') and is_x64,
         unix=plat.startswith(('linux', 'darwin')),
         os=os,
         environ=os.environ,
     )
     if selectors is None:
         selectors = []
-    selector_dict = dict((selector, selector in selectors) for selector in custom_selectors)
+    selector_dict = dict((s, s in selectors) for s in custom_selectors)
     d.update(selector_dict)
     return d
 
--- a/conda_env/env.py
+++ b/conda_env/env.py
@@ -4,6 +4,7 @@ from copy import copy
 import os
 import re
 import sys
+import platform
 
 # TODO This should never have to import from conda.cli
 from conda.cli import common
@@ -60,27 +61,51 @@ def from_yaml(yamlstr, **kwargs):
     return Environment(**data)
 
 
+def custom_selectors_from_yaml(yamlstr):
+    """Load and return a list of custom selectors from a given ``yaml string``"""
+    data = yaml.load(yamlstr)
+    if 'selectors' in data:
+        return data['selectors']
+    return []
+
+
 def from_file(filename, selectors=None):
     if not os.path.exists(filename):
         raise exceptions.EnvironmentFileNotFound(filename)
     with open(filename, 'r') as fp:
-        filtered_yamlstr = select_lines(fp.read(), filename, selectors)
-        return from_yaml(filtered_yamlstr, filename=filename)
+        yamlstr = fp.read()
+        custom_selectors = custom_selectors_from_yaml(yamlstr)
+        yamlstr = select_lines(yamlstr, filename, custom_selectors, selectors=selectors)
+        return from_yaml(yamlstr, filename=filename)
 
 
-def ns_cfg(selectors=None):
-    d = dict()
-    if selectors and len(selectors) > 0:
-        selector_dict = dict((selector, True) for selector in selectors)
-        d.update(selector_dict)
+def ns_cfg(custom_selectors, selectors=None):
+    plat = sys.platform
+    arch = platform.architecture()
+    d = dict(
+        linux=plat.startswith('linux'),
+        linux32=plat.startswith('linux') and '32' in arch,
+        linux64=plat.startswith('linux') and '64' in arch,
+        osx=plat.startswith('darwin'),
+        win=plat.startswith('win32'),
+        win32=plat.startswith('win32') and '32' in arch,
+        win64=plat.startswith('win32') and '64' in arch,
+        unix=plat.startswith(('linux', 'darwin')),
+        os=os,
+        environ=os.environ,
+    )
+    if selectors is None:
+        selectors = []
+    selector_dict = dict((selector, selector in selectors) for selector in custom_selectors)
+    d.update(selector_dict)
     return d
 
 
 sel_pat = re.compile(r'(.+?)\s*(#.*)?\[(.+)\](?(2).*)$')
-def select_lines(yamlstr, filename, selectors=None):
+def select_lines(yamlstr, filename, custom_selectors, selectors=None):
     if selectors and len(selectors) == 1 and selectors[0] == "all":
-        return yamlstr
-    namespace = ns_cfg(selectors)
+        selectors = custom_selectors
+    namespace = ns_cfg(custom_selectors, selectors)
     lines = []
     orig_lines = yamlstr.splitlines()
     for i, line in enumerate(orig_lines):
@@ -136,7 +161,7 @@ class Dependencies(OrderedDict):
 
 class Environment(object):
     def __init__(self, name=None, filename=None, channels=None,
-                 dependencies=None):
+                 dependencies=None, selectors=None):
         self.name = name
         self.filename = filename
         self.dependencies = Dependencies(dependencies)
@@ -145,12 +170,18 @@ class Environment(object):
             channels = []
         self.channels = channels
 
+        if selectors is None:
+            selectors = []
+        self.selectors = selectors
+
     def to_dict(self):
         d = yaml.dict([('name', self.name)])
         if self.channels:
             d['channels'] = self.channels
         if self.dependencies:
             d['dependencies'] = self.dependencies.raw
+        if self.selectors:
+            d['selectors'] = self.selectors
         return d
 
     def to_yaml(self, stream=None):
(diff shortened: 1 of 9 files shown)


nicoddemus:
Hmmm can I use selectors system based selectors like `win` or `linux`?

nicoddemus:
I see that you removed them in ba2e106... can you elaborate why they complicate `--select=all`?

Platform specific selectors would be a **huge** win for us, as we would not need to use Jinja preprocessing at all:

```yml
name: esss_qt10

includes:
  - {{ root }}/../plugins10core/environment.yml

dependencies:
  {% if os.environ.get('ESSS_DEBUG', '0') not in ['0', 'cpp'] %}
  - debug
  {% endif %}
  - unicodecsv==0.9.4
  {% if os.sys.platform != 'win32' %}
  - pillow==2.6.1
  {% else %}
  - pillow==2.1.0 # slightly different version copied from dist
  {% endif %}

  {% if os.environ.get('ESSS_QT_VER', '4') == '4' %}
  - pyqt==4.11.1
  - pytest-qt==1.3.0
  {% else %}
  - pyqt3==3.18.1
  {% endif %}
```

With platform-specific selectors we could us instead:

```yml
name: esss_qt10

includes:
  - {{ root }}/../plugins10core/environment.yml

dependencies:
  - debug  # [debug]
  - unicodecsv==0.9.4  
  - pillow==2.6.1  # [linux]  
  - pillow==2.1.0  # [win]
  
  - pyqt==4.11.1      # [qt4]
  - pytest-qt==1.3.0  # [qt4 and test]  
  - pyqt3==3.18.1     # [qt3]
```

Which be used with `--select test` and/or `--select debug`. The latter one looks **much** cleaner.

nicoddemus:
cc @tadeu

Korijn:
In the case of select all you have to parse the expressions and set the unknown variables to True. You can't just return everything as is brcause that would ignore the platform selectors...

Korijn:
Maybe I could catch the NameError, set the missing variable to True and keep restarting the algorithm untill no more NameErrors occur when --select all is specified...

nicoddemus:
What if we explicitly add selectors (other than platform specific ones) to the `environment.yml` file? 

```yml
name: selectors
selectors:
  - test
  - docs
dependencies:
  - python
  - pytest  # [test]
  - sphinx  # [docs]
  - doctest  # [test and docs]
```

This makes it explicit which selectors are available for use, and also allows us to implement a `list-selectors` command which lists the selectors from a `environment.yml` file.

```
$ conda env list-selectors
test
docs
```

Which is useful so other tools can query which selectors are available without resorting to parse the yaml file.

That also simplifies the implementation in the sense that it can do a stricter error checking when user passes an unknown selector in the command line.

Korijn:
Sounds like a plan to me!

nicoddemus:
Nice! :+1: 

Just to be clear, you plan to update the PR with the explicit selectors?

Korijn:
Yes, exactly.

tadeu:
This looks very good! I think that line selectors are better than `jinja2`, the file looks cleaner, and it is more simple to parse externally (less dependencies, no code execution).

For example, I'd like to make a tool to list the versions only for `compile` dependencies, so that a recompilation is triggered only when a `compile` dependency is changed.

Korijn:
Would that be an external tool or also part of a conda-env command?

nicoddemus:
@Korijn, @tadeu means an external tool. :smile:

tadeu:
At first I thought about an external tool, but it could be a parameter to `conda env export`, such as `conda env create --select compile`.

The only problem is that there is an intersection between `conda env export` and `conda list` commands, and the latter is outside `conda-env`.

nicoddemus:
> For example, I'd like to make a tool to list the versions only for compile dependencies, so that a recompilation is triggered only when a compile dependency is changed.

@tadeu great, I had a similar idea.

I think we can add also add a `conda env preprocess` command which outputs the full processed file with selectors stripped:

```
$ conda env preprocess --select test
name: selectors
selectors:
  - test
  - docs
dependencies:
  - python
  - pytest

$ conda env preprocess --select docs
name: selectors
selectors:
  - test
  - docs
dependencies:
  - python
  - sphinx
```

this way external tools can make use of the complete dependency information without duplicating the selector logic, and also taking in account platform specific selectors. This idea is out of scope as far as this PR is concerned though.

tadeu:
@nicoddemus I like that (I agree, we are getting "out-of-topic", but these are good ideas that are related to the one in this PR).

It would be good to have a common preprocessing code, because it may be further improved. For example, sometimes lists may become empty if all of its items are "unselected", and the YAML may become malformed. The preprocessor could remove the entire list in this case.

Hmm, isn't `select_lines` already implemented in `conda` by the way?

nicoddemus:
> Hmm, isn't select_lines already implemented in conda by the way?

I think it is implemented in `conda-build`.

Korijn:
It is, but I am not sure if importing it is such a good idea.. what if conda's interface changes?

Korijn:
Also, I think it depends on another function that in turn has incompatible dependencies on conda build.

Korijn:
Updated this to accept custom selectors. Please look at the enclosed test case. Also added platform selectors back in. Apologies for the delay!

@nicoddemus @msarahan Please review?

tests/test_env.py
```diff
@@ -43,6 +43,42 @@ def test_with_pip(self):
         self.assert_('foo' in e.dependencies['pip'])
         self.assert_('baz' in e.dependencies['pip'])
 
+    def test_with_selectors(self):
```
nicoddemus: Could you also check one of the platform selectors in this test? I think adding:

```
  - pywin32 # [win]
```

And adding an extra test like this:

```python
if sys.platform.startswith('win'):
    self.assert_('pywin32' in e.dependencies['conda'])
else:
    self.assert_('pywin32' not in e.dependencies['conda'])
```

would be enough

conda_env/env.py
```diff
@@ -58,11 +60,81 @@ def from_yaml(yamlstr, **kwargs):
...
+        environ=os.environ,
+    )
+    if selectors is None:
+        selectors = []
+    selector_dict = dict((s, s in selectors) for s in custom_selectors)
+    d.update(selector_dict)
+    return d
+
+
+sel_pat = re.compile(r'(.+?)\s*(#.*)?\[(.+)\](?(2).*)$')
```
nicoddemus: I have some questions about the regular expression:

* Why is the `#` optional? regex:  `(#.*)?`
* Why matching again against group `2` at the end of the regex?  regex: `(?(2).*)`

nicoddemus:
@Korijn thanks a lot for continuing to pursue this! :smile:

conda_env/env.py
Korijn: I'll be honest here: I ripped it from conda-build, assuming that it's probably safe since it's been exposed to a lot of usage already. I am an experienced regex user, but this is a mystery for me as well. Perhaps there is some use case for this that I'm not aware of...

tests/test_env.py
Korijn: Sure, I'll also add one for architecture.

conda_env/env.py
Korijn: I looked up the changeset that introduced the regex, and was surprised to learn that these selectors originally weren't embedded in comments... so I guess it's a bit of legacy?

https://github.com/conda/conda-build/commit/ccb5d729683855a037500b11c654a9abb467fbc1

Korijn: Guess what, you can/could use selectors without comments, probably before they switched to pyyaml? https://github.com/conda/conda-recipes/blob/master/matplotlib/meta.yaml

Korijn: I'll adjust the regex since we don't need to support such legacy cases.

nicoddemus: Yeah, @tadeu mentioned it to me in person. I agree with you that it seems to be supported only for legacy reasons.

Since this is a new feature, I would rather to be a little more strict and always require `#` here, but that's just my opinion.

groutr: Actually, I'd be in favor of the opposite (always require selector outside of comment).  We shouldn't expect comments to survive for very long in a .yaml file (especially any file read with pyyaml).  Requiring selectors to be in comments makes the use case of automatically modifying recipes via software almost impossible to do (since the very first bit of information thrown away are the comments).  The selectors were originally (and still are, I believe) expected to be the last thing on a line (whether in a comment or not).

msarahan: But then we get back to whether these files are actually parsable yaml, right? What makes the files most standard compatible and least lossy?

Korijn: Personally, I think the approach should be compatible with what conda-build users are already familiar with. Inconsistent syntax/API is incredibly confusing.

In fact, I fell for that trap this week on appveyor, where you can expand environment variables with windows %syntax% everywhere, except for the artifacts $(section).

I also disagree with "almost impossible". The code in this PR accomplishes it, and it's not hard at all.

On another note, the `(?(2).*)$` part of the regex is a conditional part: if group 2 (`(#.*)?`) matched, additional characters are allowed before the end of the string. In other words, *if* the selector is not in a comment, additional characters are *not* allowed after the selector.

Korijn: I think there's no harm in continuing to support both formats, to keep both parties in this discussion happy. So, to summarize, this is now considered valid:

```
name: myenv
dependencies:
  - python
  - pywin32  # [win] we need this on windows
  - py2exe  [win]
```

And this is invalid:

```
name: myenv
dependencies:
  - python
  - pywin32  [win] we need this on windows
```

tests/test_env.py
Korijn: Done!

Korijn:
I think we're good to go here.

nicoddemus:
:+1:

Korijn:
@msarahan can this be reviewed and possibly be merged in? I've put a lot of time and effort into this PR so it'll satisfy a large number of users. Hoping to get it in soon.

tadeu:
Ping @kalefranz

msarahan:
LGTM. Nice improvements.  Leaving it to @kalefranz for final say.

Korijn:
One more ping @msarahan and @kalefranz for this, I'd really like to get this in! :)

msarahan:
Thanks @Korijn, sorry for the delay.  We'll try to get this into a release soon.

Issue #218, opened 2016-01-01
endlocal statement in activate.bat prevents environment variables being propogates


stephenakearns:
_From @gillins on December 15, 2015 3:3_

Just testing the latest gdal build. There is a gdal-activate.bat that gets properly called to set %GDAL_DATA% but this does not get returned to the user's environment. I think this is due to the endlocal statement preventing set variables getting returned to the caller.

This is easy to test by trying to activate an environment with gdal in it and testing the value to %GDAL_DATA%

_Copied from original issue: Anaconda-Server/support#4_

stephenakearns:
_From @srossross on December 18, 2015 17:20_

@csoja when you have a moment can you make an update on this issue?

stephenakearns:
Hi @gillins - to clarify, you've created an environment with gdal installed and then when you go to conda env source activate you get this error condition where the env variable isn't being set?  cc: @groutr

stephenakearns:
_From @gillins on January 1, 2016 4:30_

Correct.

I can also confirm that putting an endlocal statement in gdal-activate.bat didn't seem to help.

gillins:
Any progress with this? Is a serious problem for advanced GDAL users on Windows...

msarahan:
Please see https://github.com/conda/conda-env/pull/219

As a stopgap, you can replace your activate.bat and deactivate.bat with the modified files in that PR.

Ping @groutr @kalefranz

gillins:
Having to patch installed files isn't a great introduction to Conda for new users who need GDAL. Surely something better can be done here to have it work on a clean install even if there are tweaks to the details later?

msarahan:
I agree. Ping @kalefranz - can we have a conda environment release soon? We need to remove the activate scripts from it, since they live in conda now.  I will submit a PR in a bit.

I wrote a test case specifically for this, so I'm pretty sure it will work after PR 1727.

gillins:
I see #219 has been closed. The current version of activate.bat doesn't seem to allow anything other than PATH and PROMPT to be saved. Has there been a change of plan on how to resolve this?

msarahan:
Conda PR 1727 has been merged and will be in the coming conda 4.1 release.  The test for this behavior is at https://github.com/conda/conda/blob/master/tests/test_activate.py#L712

If you see any deficiency in that test, please let us know how to improve it.

msarahan:
Closing.  This is fixed in activate scripts that now live in conda, not conda env.

PR #219, opened 2016-01-01
move activate.d and deactivate.d script running into ENDLOCAL

msarahan:
This makes it possible for such scripts to set environment variables.  

Closes https://github.com/conda/conda/issues/1916
Closes https://github.com/conda/conda-env/issues/218

I'm not completely certain that this is the right way to do this.  This potentially changes the values of variables that any package activate/deactivate scripts would see.  Unforeseen consequences are possible, but I don't see a way to make this work otherwise.  Shifting to all global variables has more potential for problems, in my opinion.

Ping @groutr @csoja

### DIFF
--- a/bin/activate.bat
+++ b/bin/activate.bat
@@ -55,13 +55,6 @@ IF "%CONDA_DEFAULT_ENV%" == "" GOTO skipdeactivate
     REM This search/replace removes the previous env from the path
     ECHO Deactivating environment "%CONDA_DEFAULT_ENV%"...
 
-    REM Run any deactivate scripts
-    IF NOT EXIST "%CONDA_DEFAULT_ENV%\etc\conda\deactivate.d" GOTO nodeactivate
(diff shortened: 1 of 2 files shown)


msarahan:
Pinging @groutr again.  This has come up a couple of times recently.

tadeu:
@kalefranz A lot of `activate.d` scripts need to specify new environment variables (actually this seems like one one of the main reasons of the existence of `activate.d` scripts). It seems that @msarahan implementation would work (another option would be to not use `setlocal`/`endlocal`). Why was this PR closed?

cc @groutr

msarahan:
@tadeu - don't worry, this is still happening.  It's in master, and I have tests for it.  It will be part of the conda 4.1 release.

msarahan:
master of conda, that is, not conda-env.  Activate scripts are over there now.

tadeu:
Aw, thanks!
Waiting for 4.1 plus the new synced conda-env release then :)

PR #224, opened 2016-01-26
Check return code when calling pip.

### LINKED ISSUE #223: conda env update/create return success even if pip fails
When installing pip packages into a conda environment using an environment.yml file the conda env update/create process will return with success even if pip fails.

Normally this isn't too much of an issue as the pip error message is in the output but when using an automated deployment (e.g. with ansible) the process' return code is important.

An example environment.yml file, the output using the file and conda list/conda info output are here:

https://gist.github.com/dmkent/a18c4fbbc80132d52f24

Should just be a matter of checking the return code of the subprocess in installers/pip.py.

I've got a patch near ready for this and will submit a PR.

dmkent:
Closes #223 

This checks pip's return code and fails the install if pip failed. Adds some mocked tests to check this behaviour.

### DIFF
--- a/conda_env/installers/pip.py
+++ b/conda_env/installers/pip.py
@@ -1,6 +1,7 @@
 from __future__ import absolute_import
 import subprocess
 
+from conda.cli import common
 from conda.pip import pip_args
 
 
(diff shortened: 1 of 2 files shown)


msarahan:
Thanks!

PR #229, opened 2016-03-09
Add a mechanism to let an environment disable the default channels

mwiebe:
Presently the `conda env` command has no way to create environments that are isolated from the anaconda default channels. In `conda` itself, this is done via the `--override-channels` option, but adding that option doesn't bundle that with the environment itself. It makes more sense for the creator of the environment to make this choice than the person running the `conda env` command, so it feels more appropriate to add it to the .yaml file.

This adds the ability to disable the default channels when defining an environment .yaml file by adding a `nodefaults` item to the channels list. This is inspired by the `defaults` in `.condarc`, just with the inverse meaning.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -79,6 +79,11 @@ in your local package cache.
 You can explicitly provide an environment spec file using ``-f`` or ``--file``
 and the name of the file you would like to use.
 
+The default channels can be excluded by adding ``nodefaults`` to the list of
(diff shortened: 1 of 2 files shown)


mwiebe:
Ping @msarahan @kalefranz to get eyeballs on this. (Looking at the most recently merged PR, this appears to be the thing to do. ;)

mwiebe:
I've rebased this and fixed up the merge conflict.

PR #244, opened 2016-05-31
Conda envs outside of default path

### LINKED ISSUE #236: Feature request: conda-env should be able to create env outside default path
With conda create I can use --prefix to specify installation path. I would like to have the same ability for conda-env. 

Reason is: we are planing to use conda-env in an CI setup where we want to create a custom environment for every build. It will be much easier to manage this if the env lies together with the project instead of all envs in anacondas install directory.

jerowe:
Hello,

I have made a few changes for the ability to add conda-env envs outside of the default path.
 
conda env create -p /home/user/software/deathstar
of
conda env create -n deathstar

I also made a few changes to the attach code to reflect this, and to add the prefix to the metadata added by conda attach.

I'm not sure how to go about listing environments. I've noticed this is a feature missing from the main conda, when doing conda create -p /home/user/software/deathstar, conda list would not show this. I am thinking perhaps on install a ~/.conda-env.rc is created with full paths to the envs, but I wanted to get some input before I went and implemented this.

Thanks so much for the great project!

Closes #236

### DIFF
--- a/conda_env/cli/main_attach.py
+++ b/conda_env/cli/main_attach.py
@@ -32,6 +32,13 @@ def configure_parser(sub_parsers):
         default=None
     )
     group.add_argument(
+        '-p', "--prefix",
+        action="store",
+        help="Full path to environment prefix",
+        metavar='PATH',
+        default=None
+    )
+    group.add_argument(
         '-r', '--remote',
         action='store',
         help='remote environment definition',
@@ -61,7 +68,11 @@ def configure_parser(sub_parsers):
 
 
 def execute(args, parser):
-    if args.name is not None:
+    # if args.name is not None:
+        # prefix = common.get_prefix(args)
+        # content = from_environment(args.name, prefix, no_builds=args.no_builds).to_dict()
+
+    if args.prefix is None:
         prefix = common.get_prefix(args)
         content = from_environment(args.name, prefix, no_builds=args.no_builds).to_dict()
     else:
--- a/conda_env/cli/main_attach.py
+++ b/conda_env/cli/main_attach.py
@@ -1,5 +1,5 @@
 from argparse import RawDescriptionHelpFormatter
-from ..utils.notebooks import current_env, Notebook
+from ..utils.notebooks import Notebook
 from conda.cli import common
 from ..env import from_environment
 
@@ -32,16 +32,16 @@ def configure_parser(sub_parsers):
         default=None
     )
     group.add_argument(
+        '-r', '--remote',
+        action='store',
+        help='remote environment definition',
+        default=None
+    )
+    p.add_argument(
         '-p', "--prefix",
         action="store",
         help="Full path to environment prefix",
         metavar='PATH',
-        default=None
-    )
-    group.add_argument(
-        '-r', '--remote',
-        action='store',
-        help='remote environment definition',
         default=None
     )
     p.add_argument(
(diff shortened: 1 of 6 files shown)


conda_env/cli/main_create.py
```diff
@@ -42,13 +42,9 @@ def configure_parser(sub_parsers):
         help='environment definition file (default: environment.yml)',
         default='environment.yml',
     )
-    p.add_argument(
```
msarahan: Why are you killing -n/--name?  Are you sure that no one else uses it?

conda_env/cli/main_create.py
```diff
@@ -79,19 +75,23 @@ def configure_parser(sub_parsers):
...
         spec = specs.detect(name=name, filename=args.file,
                             directory=os.getcwd(), selectors=args.select)
         env = spec.environment
 
         # FIXME conda code currently requires args to have a name or prefix
-        if args.name is None:
+        if args.prefix is None:
             args.name = env.name
 
+        #print("Args name is {}".format(args.name))
```
msarahan: Please remove extraneous comments

conda_env/cli/main_attach.py
```diff
@@ -61,8 +68,18 @@ def configure_parser(sub_parsers):
 
 
 def execute(args, parser):
-    if args.name is not None:
+    # if args.name is not None:
+
+        # prefix = common.get_prefix(args)
```
msarahan: Please remove extraneous comments

msarahan:
@jerowe thanks for your work on this.

Regarding your point about ```conda list``` - the really important thing to keep in mind is that conda environments are (currently) independent of a given root miniconda/anaconda installation.  This is really nice, because if the root installation is ever corrupted, then the environments can just be moved/copied somewhere else.  Your idea of a new RC file is a good implementation that would keep this.  I'd prefer, however, to use something that already exists.  

What do you think about adding/modifying the CONDA_ENVS_DIR environment variable whenever the -p option is used, so that the parent folder of these new environments is added there?

conda_env/cli/main_create.py
nicoddemus: We use `-n` extensively on our CI setup.

jerowe: I didn't kill -n. Its supplied by conda.common.add_parser_prefix

conda_env/cli/main_attach.py
jerowe: Done

jerowe:
@msarahan - I think using the envs_dirs of the ~/.condarc is an interesting idea. I'm not sure what the gain is over just using the prefix/name provided by conda.common.add_prefix_parser, but if you would prefer it done that way I'm happy to create a pull request in that direction.

I'm not overly concerned with the conda list function, but I wasn't sure if its a concern to anyone else. ;-) I think that if anyone is concerned with prefixes they are using a separate system (environment modules on an hpc cluster, probably) to track those. You don't get envs outside of the root_dir when using the main conda list either.

jerowe:
Hi @msarahan an, any updates? I'd love to have this feature in conda env. My whole infrastructure would be config files, and I just can't even express how happy that would make me. ;-)

msarahan:
Thanks @jerowe - @kalefranz any estimate on next release here?

jerowe:
Any news on the next release? Any chance it would be out by August?

msarahan:
Sorry @jerowe, pinging @kalefranz for when he might cut the next release.

conda_env/cli/main_create.py
jseabold, line 45: Are we positive that this isn't the cause of #254? Not sure there was proper test coverage here.

Issue #248, opened 2016-06-15
As of conda 4.1, conda env remove fails with AttributeError: module 'conda.plan' has no attribute 'PREFIX'


kalefranz:
@ronert reports https://github.com/conda/conda/issues/2679:

I tried this on three separate machines with conda 4.1.0. Every machine fails with `AttributeError: module 'conda.plan' has no attribute 'PREFIX'` when running `conda env remove -n environment_name` or `conda env remove -p environment_path`. This also happens on my Travis environment now causing my builds to fail. Any ideas?

kalefranz:
A current alternate solution is

    conda remove -n environment_name --all

Korijn:
I can't provide details (on my phone) but this feature of conda env has never worked for me since I became a user 6 months ago, and neither has `conda env create --force`. So, IMHO, I don't think this is a regression type of bug.

kalefranz:
Fixed with #251.

PR #250, opened 2016-06-16
Pull PREFIX/UNLINK from their proper place

mcg1969:
Fixes https://github.com/conda/conda-env/issues/248, https://github.com/conda/conda/issues/2679


mcg1969:
NOTE: the right thing to do here is forward this along to the `conda remove --all` logic in `conda/cli/main_remove.py`.

mcg1969:
For comparison, see PR #251.

PR #251, opened 2016-06-16
Just let conda/cli/main_remove do the remove work

mcg1969:
Even better than #250. Just forward `conda env remove` along to the `conda remove --all` logic.

### DIFF
--- a/conda_env/cli/main_remove.py
+++ b/conda_env/cli/main_remove.py
@@ -1,7 +1,5 @@
 from __future__ import absolute_import, print_function
-from argparse import RawDescriptionHelpFormatter
-from os.path import basename
-import textwrap
+from argparse import RawDescriptionHelpFormatter, Namespace
 
 from conda.cli import common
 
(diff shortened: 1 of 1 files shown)


Korijn:
Is there a unit test for this logic?

mcg1969:
We do exercise `conda remove` in integration testing...

mcg1969:
I'm not familiar with the `conda-env` test suite but a cursory look suggests that `conda env remove` is not exercised, even before this PR.

Korijn:
Ok! There is another code path `conda env create --force` that removes an environment (using `conda.install.rm_rf`) if it exists before creating it anew. It's currently broken (at least on Windows), and by hooking it up to this code it would probably be fixed. Would you consider including that change in this PR? It should be fairly simple.

mcg1969:
I'm not sure you can do that here, since that requires a less interactive result.

Korijn:
You could pass `yes: True` and possibly `quiet: True` as well along in that case to bypass interactivity, perhaps?

mcg1969:
That certainly seems reasonable but I must defer to the primary conda-env maintainer here. I think I would rather see `rm_rf` fixed, however, because it is used elsewhere in `conda`.

Korijn:
Cc @msaharan @kalefranz ?

mcg1969:
No need to ping them. They will both be looking at this soon. You and I have been having a conversation in the middle of the night / early morning our time.

Korijn:
Oh, OK! Sorry, I'm not really aware of the structure of your team etc. :)

mcg1969:
Oh, no problem! Yesterday was a very eventful day for us all, so I have no doubt they're getting some much-needed sleep :-)

kalefranz:
This one.

PR #253, opened 2016-06-16
remove selectors

kalefranz:
@mcg1969 I win.


Korijn:
Well, this is unexpected! Might I ask why? 

I contributed this code and invested a lot of effort into this. If there's any issues that you are experiencing I'd love to see if there is an alternative solution to just ripping it out...

kalefranz:
CC @pzwang

pzwang:
Yep, please discussion offline.

kalefranz:
@Korijn sent an email to your gmail address.  Happy to discuss further.

ocefpaf:
@kalefranz I just found this and I am curious why this was removed too. I understand that having selectors in the `environment.yml` can lead to poor reproducibility as envs are not guaranteed to be the same under different platforms. However, it would be nice to be able to add some platform specific packages in envs that are not really related to the reproducibility. For example, adding `console_shortcut` for Windows envs.

Issue #254, opened 2016-06-17
No longer able to override file's name using --name command line argument


joshowen:
Previously we could override the name from an environment file using `--name` in the CLI, but this is no longer respected after the latest release.  This was very useful for heroku buildpacks.

dharhas:
bump. I'm seeing this too. The -n/--name is no longer doing anything.

nehalecky:
bump, same issue. Any workarounds in the interim?

nehalecky:
FWIW, I can get around this by:

```bash
conda create --yes --quiet --name=my-env python=3.5.1
conda update --name=my-env --quiet --yes --all
conda env update  --quiet --name=my-env --file=dependencies.yml
```

jseabold:
What version of conda are you using for this workaround? What version of conda created the environment.yml file?

nehalecky:
@jseabold, the `environment.yml` isn't create by conda, it's crafted by hand. :)

Here's what it looks like!
```
name: coral
dependencies:
# core coral dependencies (must be installed)
  - python>=3.5
  - pip
  - pyyaml
  - pandas
  - numexpr
  - coverage
  - pytest
  - pytest-cov
  - pip:
    # core dependencies
    - findspark
    - bumpversion
    - invoke==0.12.2
    # analytics dependencies
    - coveralls
```


Version info, below:

```
❯ conda info          
Current conda install:

             platform : osx-64
        conda version : 4.1.6
    conda-env version : 2.5.1
  conda-build version : 1.21.5
       python version : 3.5.2.final.0
     requests version : 2.10.0
     root environment : /Users/nehalecky/miniconda3  (writable)
  default environment : /Users/nehalecky/miniconda3
     envs directories : /Users/nehalecky/miniconda3/envs
        package cache : /Users/nehalecky/miniconda3/pkgs
         channel URLs : https://repo.continuum.io/pkgs/free/osx-64/
                        https://repo.continuum.io/pkgs/free/noarch/
                        https://repo.continuum.io/pkgs/pro/osx-64/
                        https://repo.continuum.io/pkgs/pro/noarch/
          config file : /Users/nehalecky/.condarc
         offline mode : False
    is foreign system : False
```

HTH, let me know if you need any other info! Thanks!

jseabold:
[...]

PR #261, opened 2016-07-11
Fix environment parsing with channels

### LINKED ISSUE #257: environment.yml ill-parsed with conda 4.1.3?
```bash
saket@launchpad : ~/
[0] % conda -V
conda 4.1.3
```

It looks like 4.1.3. introduced putting [`channels' key](https://gist.github.com/saketkc/c43b6f2e6a90fa7bdbc8d0cbc3a43d6d#file-multiz-yml-L4) in `environment.yml`, besides appending the channel name for the package in `dependencies'. E.g. [`bioconda::ucsc-chainantirepeat=332=0`](https://gist.github.com/saketkc/c43b6f2e6a90fa7bdbc8d0cbc3a43d6d#file-multiz-yml-L56)

Here is my [environment.yml](https://gist.github.com/saketkc/c43b6f2e6a90fa7bdbc8d0cbc3a43d6d)

Trying to create new environment:

```
[1] % conda env create -f environment.yml 
Using Anaconda Cloud api site https://api.anaconda.org
Error: Can't process without a name
environment.yml is not a valid yaml file.
Environment with requierements.txt file needs a name
(multiz) 
```

Also bumped into #254

jseabold:
Since I've been blocked on this. This fixes the second issue I mentioned in #257 but not the first. There is a new error returned by the first that is related to https://github.com/conda/conda/issues/2995 and #254 and is addressed in #262.

### DIFF
--- a/conda_env/env.py
+++ b/conda_env/env.py
@@ -51,7 +51,9 @@ def from_environment(name, prefix, no_builds=False):
     if len(pip_pkgs) > 0:
         dependencies.append({'pip': ['=='.join(a.rsplit('-', 2)[:2]) for a in pip_pkgs]})
 
-    channels = config.get_rc_urls()
+    # conda uses ruamel_yaml which returns a ruamel_yaml.comments.CommentedSeq
+    # this doesn't dump correctly using pyyaml
+    channels = list(config.get_rc_urls())
 
     return Environment(name=name, dependencies=dependencies, channels=channels, prefix=prefix)
 


jseabold:
This should also fix the name being overwritten now. This could probably use some tests.  **Edit**: name parsing is fixed in #262 with a test.

jseabold:
Given that I can't [round-trip the yaml with ruamel_yaml](https://github.com/conda/conda/issues/2800#issuecomment-232086259), unless @kalefranz wants to fix this, I'm going to leave it and see about a test.

conda_env/env.py
```diff
@@ -51,7 +51,9 @@ def from_environment(name, prefix, no_builds=False):
     if len(pip_pkgs) > 0:
         dependencies.append({'pip': ['=='.join(a.rsplit('-', 2)[:2]) for a in pip_pkgs]})
 
-    channels = config.get_rc_urls()
+    # conda uses ruamel_yaml which returns a ruamel_yaml.comments.CommentedSeq
+    # this doesn't dump correctly using pyyaml
```
kalefranz, line 55: > this doesn't dump correctly using pyyaml

I think this is the key issue.  There's no problem with this PR.  But I think we also need to force `ruamel_yaml` or `ruamel.yaml` instead of `pyyaml` for conda-env now that we're using it with conda.

kalefranz:
FWIW, this whole conda-env repo has been sucked up into the conda repo.  We'll soon be deprecating the whole repo.  But going to go ahead with these patches here for now.

conda_env/env.py
kalefranz, line 55: Well, I was _going_ to make a patch that replaced pyyaml with ruamel_yaml, but conda-env apparently has a custom wrapper around pyyaml, so it's not a straight-forward replacement.  Will take a bit more time than just switching out the library.

If you or someone else wants to do that, please be my guest.

Today and tomorrow I'm focused on getting conda canary 4.2.0 out.

Issue #270, opened 2016-08-24
Request  - Versioning of Conda Envs


jerowe:
Hello,

I would like to be able to version environments the way I do packages.

When I use anaconda to upload a conda env it creates a version number from the current datetime. Instead I would like something 1.0, 1.1, etc.

With uploading the relevant code is in 'conda_env.utils.uploader', and there is a function version. Currently this returns the relevant datetime. Instead I would like it to look at a specified version in the environment.yml file, and if that's not found go about creating its usual version.

I can, btw, create an environment.yml file with a version field, and upload that to conda env. The version field doesn't actually do anything, but its there and doesn't get any errors. When I try to install that with conda env create user/package/2016.08.23.1114, I get the following stack trace.

`
Traceback (most recent call last):
  File "/home/jillian/.local/easybuild/software/anaconda/3-4.0.0/bin/conda-env", line 6, in <module>
    sys.exit(conda_env.cli.main.main())
  File "/home/jillian/.local/easybuild/software/anaconda/3-4.0.0/lib/python3.5/site-packages/conda_env/cli/main.py", line 68, in main
    return args_func(args, parser)
  File "/home/jillian/.local/easybuild/software/anaconda/3-4.0.0/lib/python3.5/site-packages/conda/cli/main.py", line 130, in args_func
    exit_code = args.func(args, p)
  File "/home/jillian/.local/easybuild/software/anaconda/3-4.0.0/lib/python3.5/site-packages/conda_env/cli/main_create.py", line 77, in execute
    env = spec.environment
  File "/home/jillian/.local/easybuild/software/anaconda/3-4.0.0/lib/python3.5/site-packages/conda_env/specs/binstar.py", line 93, in environment
    return env.from_yaml(self._environment)
  File "/home/jillian/.local/easybuild/software/anaconda/3-4.0.0/lib/python3.5/site-packages/conda_env/env.py", line 65, in from_yaml
    return Environment(**data)
TypeError: __init__() got an unexpected keyword argument 'version'
`

I've poked around enough in the code that I could probably implement this, but I'm not sure if anyone else would like to have it.

jerowe:
I can, btw, do this.

```
> anaconda show jerowe/gencore_build_1.0/2016.08.24.1415       
Using Anaconda API: https://api.anaconda.org
version 2016.08.24.1415
   + environment-1.0.yml
None
```

So searching based on version is supported by anaconda.

jerowe:
[...]

PR #271, opened 2016-08-24
Add versioning to conda envs

jerowe:
[270](https://github.com/conda/conda-env/issues/270)

I would like to add user controlled versioning to conda envs. I've got the uploader part done, but I still need to do some checks for creating. Here is my initial effort.

Best,
Jillian

### DIFF
--- a/conda_env/specs/binstar.py
+++ b/conda_env/specs/binstar.py
@@ -8,7 +8,6 @@ try:
 except ImportError:
     get_binstar= None
 try:
-    from binstar_client import errors
     from binstar_client.utils import get_server_api
 except ImportError:
     get_server_api = None
@@ -31,6 +30,7 @@ class BinstarSpec(object):
     _packagename = None
     _package = None
     _file_data = None
+    _specs = None
     msg = None
 
     def __init__(self, name=None, **kwargs):
--- a/conda_env/specs/binstar.py
+++ b/conda_env/specs/binstar.py
@@ -31,6 +31,7 @@ class BinstarSpec(object):
     def __init__(self, name=None, **kwargs):
         self.name = name
         self.quiet = False
+        self.version = None
         if get_binstar is not None:
             self.binstar = get_binstar()
         else:
@@ -81,15 +82,24 @@ class BinstarSpec(object):
         :raises: EnvironmentFileNotDownloaded
         """
         if self._environment is None:
+
             versions = [{
                 'normalized': normalized_version(d['version']),
                 'original': d['version']} for d in self.file_data]
-            latest_version = max(versions, key=lambda x: x['normalized'])['original']
-            file_data = [data for data in self.package['files'] if data['version'] == latest_version]
-            req = self.binstar.download(self.username, self.packagename, latest_version, file_data[0]['basename'])
+
+            if self.version:
+                print("We have a version! {}".format(self.version))
+                file_data = [data for data in self.package['files'] if data['version'] == self.version]
+                req = self.binstar.download(self.username, self.packagename, self.version, file_data[0]['basename'])
+            else:
+                latest_version = max(versions, key=lambda x: x['normalized'])['original']
+                file_data = [data for data in self.package['files'] if data['version'] == latest_version]
+                req = self.binstar.download(self.username, self.packagename, latest_version, file_data[0]['basename'])
+
             if req is None:
                 raise EnvironmentFileNotDownloaded(self.username, self.packagename)
             self._environment = req.text
+
         return env.from_yaml(self._environment)
 
     @property
@@ -117,4 +127,18 @@ class BinstarSpec(object):
 
     def parse(self):
         """Parse environment definition handle"""
-        return self.name.split('/', 1)
+        split_name = self.name.split('/')
+
+        if len(split_name) == 3:
+            #We have a version
+            self.version = split_name[2]
+            return split_name[0:2]
+        elif len(split_name) == 2:
+            #No version should be user/package
+            return split_name
+        else:
+            #We should raise some sort of exception here
+            self.msg = "Remote definition name does not match convention. It should be user/package. Example: darth/deathstar"
+            return False
+
+        # return self.name.split('/', 1)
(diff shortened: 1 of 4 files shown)


jerowe:
You can now add a version tag to an environment.yml file, which will be correctly propogated to anaconda cloud.

In addition you can create a specific version of an environment.

```
conda env create jerowe/gencore_build_1.0/1.0    #version
conda env create jerowe/gencore_build_1.0          #gets the latest version
```

jerowe:
Hi @msarahan. You took a look at my last PR. Could you take a look at this? If not something the conda env community is interested in I completely understand, especially with some of the cool new changes like jinja templates you guys want to implement. If you could give me a general yay/nay that would be excellent.

Best,
Jillian

conda_env/specs/binstar.py
```diff
@@ -81,15 +82,24 @@ def environment(self):
...
+
             versions = [{
                 'normalized': normalized_version(d['version']),
                 'original': d['version']} for d in self.file_data]
-            latest_version = max(versions, key=lambda x: x['normalized'])['original']
-            file_data = [data for data in self.package['files'] if data['version'] == latest_version]
-            req = self.binstar.download(self.username, self.packagename, latest_version, file_data[0]['basename'])
+
+            if self.version:
+                print("We have a version! {}".format(self.version))
```
msarahan: This seems like a debugging message.  Can it be cut or turned into a log message instead?

conda_env/specs/binstar.py
```diff
@@ -117,4 +127,18 @@ def packagename(self):
...
-        return self.name.split('/', 1)
+        split_name = self.name.split('/')
+
+        if len(split_name) == 3:
+            #We have a version
+            self.version = split_name[2]
+            return split_name[0:2]
+        elif len(split_name) == 2:
+            #No version should be user/package
+            return split_name
```
msarahan: This looks like a potentially dangerous change in behavior.  You have gone from a limited number of splits to an unlimited one.  Would split('/', 2) work?

msarahan:
@jerowe I think this is a nice addition.  I have made some minor comments.  @kalefranz is in charge of conda-env, and will need to make the decision on this PR.  He may want you to put this PR against conda/conda instead, as he is trying to merge conda and conda-env.

conda_env/specs/binstar.py
jerowe: I removed it. 

Is there a logging utlity?

conda_env/specs/binstar.py
jerowe: If we have a limited number of splits potentially users could throw something in there like

```
darth/deathstar/1.0/someotherthing
```

and it would get split as 

```
darth, deathstar, 1.0/thing
```

Which is obviously wrong because the length of the split doesn't correspond to anything. There should be some sort of limit that it can't split as more than 2 (user, package) or 3 (user, package, version). 

But yes, split('/', 2) would work for either case. It just wouldn't catch errors in having too many things supplied by the user. If you prefer I will do that.

msarahan: Thanks for the clarification.  I don't have a strong opinion here, just wanted to make sure you had thought about it.

jerowe:
@msarahan, thanks for the comments!

I thought I would go into a bit of why we want this.

I worked for a genomics core closely beside the HPC department, and conda along with conda env has become a major part of our infrastructure. We have a whole workflow from researchers requesting software stacks for analyses, configuring those analyses using conda env environment.yml files, and building those environments against a docker image to check for conflicting versions. Each of these conda envs is then built as a HPC module.

Conda env is wonderful, but what would really take it to the next level is being able to version the envs. For instance, more than one researcher may be using basically the same environment, but each needs a different version of software X for whatever reason. Or there are some major updates in the software stack. These envs would be gencore_cure_cancer/1.0 and gencore_cure_cancer/1.1. 

In any case we have had great success with conda env. Thanks for the great project! ;-)

msarahan:
Yes, I understand the use case, and I agree it makes a lot of sense.  As an intermediate, you could always place your environment.yml scripts under version control, and check out tags.  I understand the barrier to entry there is higher, and you also have less direct knowledge of the environment, though.

jerowe:
@msarahan, what we're doing now is actually appending the version names to the environment names themselves. Environment names are gencore_env_1.0, gencore_env_1.1, but when the HPC module is built it is gencore_env/1.0. It is working, so its not as if its a huge hurtle. I just like the idea of versioned environments better. Environment files are already under version control, but I hadn't thought of tags. I was just hard coding everything. Things to think about!

jerowe:
@msarahan , I have a question about code change. 

According to anaconda_client the function 'get_binstar' has been deprecated in favor of 'get_server_api'. They both take the same arguments, so substituting them would be a simple change. Should I go ahead and make that change while I'm at it?

conda_env/specs/binstar.py
jerowe: So as it turns out this functionality is already implemented by 

```
binstar_client.utils.spec parse_specs
```

jerowe:
Alright, I tried to do some refactoring, but then broke the mock tests. I added back in a binstar, and will wait for feedback before I go all out with refactoring.

ETA - I'm sorry. I just can't help myself.

jerowe:
I'm not a real python developer, and I have few questions if you care to answer.

What is the point of this?

``` python
class BinstarSpec(object):
    _file_data = None
...
    @property
    def file_data(self):
        if self._file_data is None:
            self._file_data = [data for data in self.package['files'] if data['type'] == ENVIRONMENT_TYPE]
        return self._file_data
```

We have _file_data, and we have a property file data. Do I need to just go and read up on property decorators? Why not just return the [data for data STUFF] the first time around?

jerowe:
@msarahan, any word on this feature? It would be very useful for us to have!

jerowe:
@msarahan , how is it going?

Any word? We'd really like to have this feature. I would also really like to add summary and tags/labels fields, but versioning is the most important.

msarahan:
Sorry, I am out on paternity leave. Pinging @kalefranz.

At the very least, you probably need to merge this into conda instead, as conda env had been rolled into conda proper.

jerowe:
Congratulations! I have a 2 year old and a 5 year old.  They are the best.  Enjoy your time with the baby!

jerowe:
@kalefranz, What do you think?

I am very interested in versioning, summaries, and tags, but I understand if this is not something the conda community is not interested in.

Issue #241, opened 2016-05-19
Cannot export environment outside of default path


mankoff:
```
conda create -y -p ./py_foo
source activate ./py_foo
conda env export
```

Gives error: `Error: '/' not allowed in environment name: /nonstandard/path/to/py_foo`

Originally discussed here: https://groups.google.com/a/continuum.io/forum/#!topic/anaconda/8eyh0gomA4k

chemotaxis:
I can confirm that if I do the following:

```
conda create --name minimal scipy
source activate minimal
conda env export > minimal.yml
```

I get this: `Error: '/' not allowed in environment name: ~/miniconda3/envs/minimal`

and nothing is written to `minimal.yml`.

But if I do this (while `minimal` environment is still activated):
```
conda env export --name minimal > minimal.yml
```

My `minimal` environment dependencies are written to `minimal.yml`.

PR #272, opened 2016-08-26
fix for export outside of default env path

jerowe:
Fixes issue https://github.com/conda/conda-env/issues/241

```
#without a name supplied name is the basename of the prefix
conda env export -p /path/to/my_env

#Or supply a name
conda env export --path /path/to/my_env --name with_this_name
```

### DIFF
--- a/conda_env/cli/main_export.py
+++ b/conda_env/cli/main_export.py
@@ -8,6 +8,7 @@ from conda import config
 from conda.cli import common
 
 from ..env import from_environment
+from conda.config import default_prefix
 
 description = """
 Export a given environment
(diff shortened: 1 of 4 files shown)


kalefranz:
`conda-env` code moved to the `conda` proper repository a while back.  I left PRs open so people could port them over.  Obviously, I could still do better at communicating.

If you want to pull this over to the conda repo, it'll be more straight-forward to take from there :)

jerowe:
Happy to do so, but I have another PR open, and its failing. I'm not sure if its failing because of me or because of something else. Once I get that figured out (and I'd appreciate input!) I'll be happy to open the PR in conda proper ;-)

